PR #11, opened 2017-10-05
MLR-174

dsteinich:
This is a no frills happy path workflow for a single transaction ddot file. Note that update will not work until MLR-176 is deployed.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyWorkflowService.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyWorkflowService.java
@@ -1,5 +1,6 @@
 package gov.usgs.wma.mlrgateway.service;
 
+import java.time.LocalDate;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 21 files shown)


dsteinich:
Good point on GSON, I didn't bring it in because Jackson was already in the project, we aren't doing much with the JSON (yet), and I am not sure JSON is the best way to handle the reporting.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyWorkflowService.java
```diff
@@ -15,57 +14,122 @@
...
+					updateTransaction(ml.get(AGENCY_CODE), ml.get(SITE_NUMBER), json);
 				}
 			} else {
-				throw new FeignBadResponseWrapper(HttpStatus.SC_BAD_REQUEST, null, "{\"error\":{\"message\":\"Unable to determine transactionType.\"},\"data\":" + json + "}");
+				WorkflowController.addStepReport(new StepReport(VALIDATION_STEP, HttpStatus.SC_BAD_REQUEST, BAD_TRANSACTION_TYPE.replace("%json%", json), ml.get(AGENCY_CODE), ml.get(SITE_NUMBER)));
 			}
 		}
 
-		return rtn;
+		notificationClient.sendEmail("test", "rtn", "drsteini@usgs.gov");
```
dsteinich: These need to change once we have authorization.
I'm not sure yet what we should put in the message...

PR #26, opened 2017-11-02
MLR-219: Added support for communicating with the Water Auth Server over SSL

zmoore-usgs:


zmoore-usgs:
@dsteinich See my latest commit on this PR. I went back and compared your springjwt gateway branch with this and found that I missed a couple of things when I was resolving the application.yml merge conflicts. After adding those bits back my local Gateway is functioning with services deployed in CHS as well as properly Authenticating via Water Auth deployed in CHS. 

Sorry if these missing pieces are what have been causing the issues that you have been running into today; I should've gone back and compared them earlier.

PR #31, opened 2017-11-07
MLR-229 / MLR-249

zmoore-usgs:
- Fixed certain workflow step errors not properly halting the transaction
- Fixed CompleteWorkflow Add Transactions not returning a 201
- Fixed CRU errors being swallowed by the validator
- Added fetching the logged-in user's email from the security context for notifications
- Added the ability to send the notifications to a configurable list of recipients in addition to the logged-in user
- Refactored services into workflows and services
- Added services for clients that didn't already have them
- Added additional detailed error logging within services
- Updated tests


src/main/java/gov/usgs/wma/mlrgateway/controller/ExportWorkflowController.java
```diff
@@ -22,10 +21,6 @@
 @Api(tags={"Export Workflow"})
 @RestController
 public class ExportWorkflowController extends BaseController {
-
-	@Value("${temporaryNotificationEmail}")
-	private String temporaryNotificationEmail;
-
 	private ExportWorkflowService export;
 	private NotificationService notificationService;
 	public static final String COMPLETE_WORKFLOW = "Complete Export Workflow";
```
mbucknell: In lines 28 and 29, do you still need these constants here?

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -111,7 +112,8 @@ private String preValidation(Map<String, Object> ml) {
 				}
 			}
 		} catch (Exception e) {
-			//Do nothing
+			//An error occurred checking for an existing record
```
mbucknell: Should we be letting the new GlobalDefaultExceptionHandler be handling these kinds of exceptions where something unexpected happen?

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -1,13 +1,29 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import org.apache.commons.lang.StringUtils;
+import org.apache.http.HttpStatus;
+import org.apache.log4j.Logger;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.provider.OAuth2Authentication;
 
 public class BaseController {
```
dsteinich: Perhaps this should be an abstract class, with a constructor that takes the notification service, and have it as a local property. The two extending classes would then include a call to this constructor, rather than passing the notification service to the notification step.

src/main/java/gov/usgs/wma/mlrgateway/service/workflow/LegacyWorkflowService.java
```diff
@@ -0,0 +1,143 @@
...
+		for (int i = 0; i < ddots.size(); i++) {
+			Map<String, Object> ml = ddots.get(i);
+			try {
+				if (ml.containsKey(TRANSACTION_TYPE) && ml.get(TRANSACTION_TYPE) instanceof String) {					
+					if (((String) ml.get(TRANSACTION_TYPE)).contentEquals(TRANSACTION_TYPE_ADD)) {
+						ml = legacyValidatorService.doValidation(ml, true);
+						ml = transformService.transform(ml);
+						json = mlToJson(ml);
+						legacyCruService.addTransaction(ml.get(AGENCY_CODE), ml.get(SITE_NUMBER), json);
+						fileExportService.exportAdd(ml.get(AGENCY_CODE), ml.get(SITE_NUMBER), json);
```
dsteinich: You missed and import piece here and on the update side - the file export service needs to use the output of the CRU create/update. The json passed into the cru does not have all of the monitoring location properties. This causes the export to fail.

src/main/java/gov/usgs/wma/mlrgateway/controller/ExportWorkflowController.java
zmoore-usgs: Ah yeah, good catch. I'll take those out.

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
zmoore-usgs: Makes sense to me

src/main/java/gov/usgs/wma/mlrgateway/service/workflow/LegacyWorkflowService.java
zmoore-usgs: Ah whoops you're right, I completely missed that part. I'll add it back in.

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -1,13 +1,29 @@
...
+import gov.usgs.wma.mlrgateway.config.WaterAuthJwtConverter;
+import gov.usgs.wma.mlrgateway.service.NotificationService;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Arrays;
+import java.util.Map;
+import org.apache.commons.lang.StringUtils;
+import org.apache.http.HttpStatus;
+import org.apache.log4j.Logger;
```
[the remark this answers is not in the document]
zmoore-usgs: @mbucknell @dsteinich Googling around it seems that org.slf4j.Logger is the one included by default with the Spring Starter so that's probably the one we want to use? See: https://www.mkyong.com/spring-boot/spring-boot-slf4j-logging-example/

In the example project they have `import org.slf4j.Logger;`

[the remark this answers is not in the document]
dsteinich: org.slf4j.Logger is the correct one - @mbucknell I think we don't need the org.slf4j:slf4j-log4j12 that was added to the poms. It actually causes a warning on startup:
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/drsteini/maven/repository/ch/qos/logback/logback-classic/1.1.11/logback-classic-1.1.11.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/drsteini/maven/repository/org/slf4j/slf4j-log4j12/1.7.25/slf4j-log4j12-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]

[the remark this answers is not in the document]
mbucknell: I'll update the java projects next week unless someone beats me to it.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
zmoore-usgs: I might leave this one as-is so that it has the more specific error message of "An error occurred while checking for an existing record." I think that will help make it clear that the error occurred with the CRU service but not during the CRU step of the workflow.

zmoore-usgs:
@dsteinich My latest Commit should address your feedback - let me know if you see anything else.

PR #33, opened 2017-11-08
MLR-208: update the exposed port in dockerfile, update keystore password

kmschoep-usgs:
variable in docker-entrypoint.sh.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -13,11 +13,7 @@ elif [ $oauthClientSecret_file ]; then
     OAUTH_CLIENT_SECRET_VAL=`cat $oauthClientSecret_file`
 fi
 
-if [ $sslStorePassword ]; then
-    keystorePassword=$sslStorePassword
-elif [ $sslStorePassword_file ]; then 
-    keystorePassword=`cat $sslStorePassword_file`
-fi
+keystorePassword=`cat $keystore_password_path`
 
 openssl pkcs12 -export -in $token_cert_path -inkey $token_key_path -name $keystoreOAuthKey -out oauth.p12 -password pass:$keystorePassword
 openssl pkcs12 -export -in $ssl_cert_path -inkey $ssl_key_path -name tomcat -out tomcat.p12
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -27,6 +27,6 @@ keytool -importcert -file $water_auth_cert_file -keystore ssl_trust_store.jks -s
 keytool -v -importkeystore -deststorepass $keystorePassword -destkeystore ssl_trust_store.jks -deststoretype JKS -srckeystore oauth.p12 -srcstorepass $keystorePassword -srcstoretype PKCS12 --noprompt
 keytool -v -importkeystore -deststorepass $keystorePassword -destkeystore ssl_trust_store.jks -srckeystore tomcat.p12 -srcstoretype PKCS12
```
kmschoep-usgs: @zmoore-usgs , should the tomcat.p12 look more similar to the other java apps?

zmoore-usgs: Yeah probably so, although Water Auth does use its own SSL cert rather than the same one used by the rest of MLR since it is not really a part of MLR proper so the secrets/config job for Water Auth will still need to pass it in

docker-entrypoint.sh
```diff
@@ -13,20 +13,17 @@ elif [ $oauthClientSecret_file ]; then
...
+keystorePassword=`cat $keystore_password_path`
 
 openssl pkcs12 -export -in $token_cert_path -inkey $token_key_path -name $keystoreOAuthKey -out oauth.p12 -password pass:$keystorePassword
 openssl pkcs12 -export -in $ssl_cert_path -inkey $ssl_key_path -name tomcat -out tomcat.p12
 
 
 keytool -importcert -file $water_auth_cert_file -keystore ssl_trust_store.jks -storepass $keystorePassword -alias auth.nwis.usgs.gov -noprompt
 keytool -v -importkeystore -deststorepass $keystorePassword -destkeystore ssl_trust_store.jks -deststoretype JKS -srckeystore oauth.p12 -srcstorepass $keystorePassword -srcstoretype PKCS12 --noprompt
-keytool -v -importkeystore -deststorepass $keystorePassword -destkeystore ssl_trust_store.jks -srckeystore tomcat.p12 -srcstoretype PKCS12
+keytool -v -importkeystore -deststorepass $keystorePassword -destkeystore ssl_trust_store.jks -deststoretype JKS -srckeystore tomcat.p12 -srcstorepass $keystorePassword -srcstoretype PKCS12 -noprompt
```
zmoore-usgs: I think `noprompt` at the end needs 2 dashes: `--noprompt`

kmschoep-usgs: the docs for keytool are showing one dash: https://docs.oracle.com/cd/E19159-01/819-3671/ablra/index.html

zmoore-usgs: Ah gotcha looks like that works too. Not sure where I read it with 2 dashes initially!

PR #35, opened 2017-11-13
MLR-208: Updated the docker filw to use the WaterAuth cert created by…

zmoore-usgs:
… the MLR project configs job. Standardized the noprompt switches. Added additional environment variable specifications when the JAR is called.


zmoore-usgs:
@kmschoep-usgs FYI - Went ahead and merged this so I could test it

PR #37, opened 2017-11-13
MLR-208: Made the keystore location configurable to attempt to setup …

zmoore-usgs:
…paths correctly.

Inspired by the way the WaterAuth and User Interface handle the keystore location.


zmoore-usgs:
@kmschoep-usgs I'm going to go ahead and merge for testing but if you see anything here let me know. Should work alright because it was mostly just a find-replace and the removing of one -D arg.

PR #39, opened 2017-11-16
MLR-251: Update services to handle a single monitoring location returned

kmschoep-usgs:
from the CRU.


src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
```diff
@@ -60,30 +60,30 @@ public String updateTransaction(Object agencyCode, Object siteNumber, String jso
 		}
 	}
 	
-	public List<Map<String, Object>> getMonitoringLocations(Object agencyCode, Object siteNumber) {
-		List<Map<String, Object>> sites = null;
+	public Map<String, Object> getMonitoringLocations(Object agencyCode, Object siteNumber) {
```
dsteinich: We should rename this to getMonitoringLocation

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -94,8 +94,7 @@ else if (validationMessage.containsKey(LegacyValidatorClient.RESPONSE_ERROR_MESS
 	}
 
 	private String preValidation(Map<String, Object> ml) {
-		TypeReference<ArrayList<Map<String, Object>>> mapType = new TypeReference<ArrayList<Map<String, Object>>>() {};
-		ArrayList<Map<String, Object>> existingRecordList = new ArrayList<>();
+		TypeReference<Map<String, Object>> mapType = new TypeReference<Map<String, Object>>() {};
```
dsteinich: remove this - see below.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -109,11 +108,7 @@ private String preValidation(Map<String, Object> ml) {
 			int cruStatus = existingRecordResponse.getStatusCodeValue();
```
dsteinich: The code in this try/catch should be to the legacyCruService.getMonitoringLocation, rather than duplicated here (missed in the big refactor).  (Similar to what is in the ExportWorkflowService now.)

PR #41, opened 2017-11-16
MLR-255: Updated the notification format. Added the ability for the…

zmoore-usgs:
…token retrieval endpoint to be disabled based on the currently deployed tier.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
@@ -54,8 +54,7 @@ public abstract class BaseController {
 		gatewayReport.remove();
 	}
 	
-	protected int notificationStep(String subject) {
-		int status = -1;
+	protected void notificationStep(String subject) {
 		List<String> notificationRecipientList;
 		String userName = "Unknown";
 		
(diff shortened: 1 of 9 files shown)


src/main/java/gov/usgs/wma/mlrgateway/controller/UtilController.java
```diff
@@ -28,9 +29,11 @@ public UtilController() {
 			@ApiResponse(code=403, message="Forbidden")})
 	@GetMapping("/token")
 	public String getToken() {
-		if(SecurityContextHolder.getContext().getAuthentication() != null){
+		if(!environmentTier.equals(ConfigurationValues.ENVIRONMENT_PRODUCTION)) {
```
zmoore-usgs: Current behavior is to allow on DEV, QA and disabled on PROD but I can change this if QA should also not be allowed. @dsteinich ?

dsteinich: Should this be flipped to prevent NPE's? environmentTier is null checked before the other uses...
Or does the default give us an empty string to nulls checking elsewhere isn't really needed?

src/main/java/gov/usgs/wma/mlrgateway/util/ConfigurationValues.java
```diff
@@ -0,0 +1,7 @@
+package gov.usgs.wma.mlrgateway.util;
+
+public class ConfigurationValues {
+	public static String ENVIRONMENT_PRODUCTION = "production";
+	public static String ENVIRONMENT_QA = "qa";
+	public static String ENVIRONMENT_DEV = "development";
+}
```
dsteinich: Please add a private constructor to prevent this class from being instantiated.

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -53,6 +57,7 @@ public static void remove() {
 	protected int notificationStep(String subject) {
```
dsteinich: The return value of this method is never used. Could save a little complexity and have it be a void.

src/main/java/gov/usgs/wma/mlrgateway/controller/UtilController.java
zmoore-usgs: The default gives us a null so a null check should be here, I'll add one

PR #44, opened 2017-11-22
MLR-260: Added the export workflow to the UI. Updated the wording of …

zmoore-usgs:
…some buttons on the UI. Added pretty-printing for returned response JSON.

### DIFF
--- a/src/main/resources/static/index.html
+++ b/src/main/resources/static/index.html
@@ -31,23 +31,34 @@
 
 		<div class="mlr-container">
 			<div class="mlr-content">
-				<div>
-					<h3 id="upload-header" class="ddot-header">Ddot File Upload</h3>
-					<form name=ddotForm" id="ddotForm" method="POST" enctype="multipart/form-data">
-						<label for="ddotFile">Select Ddot File to Upload</label>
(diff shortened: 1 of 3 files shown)


src/main/resources/static/index.html
```diff
@@ -94,12 +105,12 @@ <h4 class="access-help">USGS Policy Information Links</h4>
 			</div><!-- /footer -->
 		</div><!-- /footer panel -->
 		
+		<!-- Application -->
+		<script src="ui.js"></script>
```
mbucknell: Why was this moved? I think it's better to load after the dependencies are loaded.

src/main/resources/static/ui.css
```diff
@@ -27,32 +27,50 @@
...
-	border: 1px solid #666;
-	padding: 5px;
-	border-radius: 5px;
-	font-size: smaller;
-	font-style: oblique;
-	color: #0030ff;
-	word-wrap: break-word;
+.mlr-action {
+	display: inline-grid;
+    margin-left: 15px;
```
mbucknell: Nit: Inconsistent indentation.

PR #53, opened 2018-09-17
Fixed a bug preventing the UI from handling error responses not conta…

zmoore-usgs:
…ining a JSON map. Added the ability to configure the session timeout time and increased the default to 12 hours from 30 mins.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -1,6 +1,8 @@
 server:
   port: ${serverPort:8443}
   use-forward-headers: true
+  session:
+    timeout: ${sessionTimeout:43200}
   ssl:
     key-store: ${keystoreLocation}
     key-store-password: ${keystorePassword}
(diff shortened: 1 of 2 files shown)


src/main/resources/application.yml
```diff
@@ -1,6 +1,8 @@
 server:
   port: ${serverPort:8443}
   use-forward-headers: true
+  session:
```
zmoore-usgs: We're on Spring Boot 1.5.8 and it looks like server.session.timeout is the proper config for this version:

https://docs.spring.io/spring-boot/docs/1.5.8.RELEASE/reference/html/common-application-properties.html

src/main/resources/static/ui.js
```diff
@@ -28,8 +28,13 @@ function generalError (response) {
 			responseText = formatJsonResponse(response.responseJSON);
 		} else if(response.hasOwnProperty("responseText")) {
 			responseText = formatJsonResponse(JSON.parse(response.responseText));
+		} else if(response.status == 401) {
+			// Got a 401 before a response map could be built, so warn of a potential expired session.
+			responseText = "It appears that your session may have expired or you are not logged in. Please refresh " + 
+				"the page and try again. If this issue persists please contact the support team.";
 		} else {
-			responseText = formatJsonResponse(response);
```
zmoore-usgs: This was what was causing the error before.

PR #57, opened 2018-10-11
Changes to support file uploads larger than the default 1MB and bette…

zmoore-usgs:
…r handling of large JSON reports in the UI.

In addition to these changes some ENVs need to be overridden in the service config ENV files:

```
SERVER_TOMCAT_MAX_HTTP_POST_SIZE=25165824 (24MB)
SPRING_HTTP_MULTIPART_MAX_FILE_SIZE=10MB
SPRING_HTTP_MULTIPART_MAX_REQUEST_SIZE=20MB
```
`SERVER_TOMCAT_MAX_HTTP_POST_SIZE` needs to be larger than `SPRING_HTTP_MULTIPART_MAX_REQUEST_SIZE`.

If files are uploaded that are larger than `SPRING_HTTP_MULTIPART_MAX_REQUEST_SIZE` the server will reject the request outright which was the reason I added:

`"Connection Error with the Gateway Service. It's possible that the file you're trying to upload greatly exceeds the maximum file upload size allowed by the server. Please try a smaller file or contact the support team."`

to the UI when the response code is 0 and the request included uploading a DDot

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/controller/GlobalDefaultExceptionHandler.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/controller/GlobalDefaultExceptionHandler.java
@@ -7,11 +7,15 @@ import java.util.Map;
 
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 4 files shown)


zmoore-usgs:
Going to go ahead and merge this

PR #60, opened 2018-11-07
MLR-328: remove 404 error status message on a successful add transaction

kmschoep-usgs:
new json example in JIRA task.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
@@ -39,7 +39,7 @@ public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
 			.cors().and()
 			.authorizeRequests()
 				.antMatchers("/swagger-resources/**", "/webjars/**", "/v2/**").permitAll()
-				.antMatchers("/info**", "/health/**", "/hystrix/**", "/hystrix.stream**", "/proxy.stream**", "/favicon.ico").permitAll()
+				.antMatchers("/info**", "/health/**", "/hystrix/**", "/hystrix.stream**", "/proxy.stream**", "/favicon.ico", "/swagger-ui.html").permitAll()
 				.anyRequest().fullyAuthenticated()
 			.and()
 				.logout().permitAll()
--- a/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
@@ -65,11 +65,14 @@ public class LegacyCruService {
 		
 		ResponseEntity<String> cruResp = legacyCruClient.getMonitoringLocation((String)agencyCode, (String)siteNumber);
 		int cruStatus = cruResp.getStatusCodeValue();
+		String preValMsg = "";
 		
 		if (cruStatus == 404) {
-			if (!isAddTransaction) {
-				BaseController.addStepReport(new StepReport(SITE_GET_STEP, cruStatus,  SITE_GET_DOES_NOT_EXIST , agencyCode, siteNumber));
+			if (isAddTransaction) {
+				cruStatus = 200;
+				preValMsg = "Duplicate agency code/site number check: ";
 			}
+			BaseController.addStepReport(new StepReport(preValMsg + SITE_GET_STEP, cruStatus,  SITE_GET_DOES_NOT_EXIST , agencyCode, siteNumber));
   		} else {
 
 			try {
--- a/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
@@ -59,7 +59,7 @@ public class LegacyCruService {
 		}
 	}
 	
-	public Map<String, Object> getMonitoringLocation(Object agencyCode, Object siteNumber) {
+	public Map<String, Object> getMonitoringLocation(Object agencyCode, Object siteNumber, boolean isAddTransaction) {
 		Map<String, Object> site = new HashMap<>();
 		ObjectMapper mapper = new ObjectMapper();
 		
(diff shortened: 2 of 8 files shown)


src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
```diff
@@ -39,7 +39,7 @@ protected void configure(HttpSecurity http) throws Exception {
 			.cors().and()
 			.authorizeRequests()
 				.antMatchers("/swagger-resources/**", "/webjars/**", "/v2/**").permitAll()
-				.antMatchers("/info**", "/health/**", "/hystrix/**", "/hystrix.stream**", "/proxy.stream**", "/favicon.ico").permitAll()
+				.antMatchers("/info**", "/health/**", "/hystrix/**", "/hystrix.stream**", "/proxy.stream**", "/favicon.ico", "/swagger-ui.html").permitAll()
```
cschroed-usgs: Good call to permit anon swagger access!

src/test/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowServiceTest.java
```diff
@@ -135,17 +136,23 @@ public void oneAddTransaction_completeWorkflow_thenReturnCreated() throws Except
...
-
+		
 		service.completeWorkflow(file);
+		
+		String mapperRet = mapper.writeValueAsString(WorkflowController.getReport());
 
 		JSONAssert.assertEquals(msg, mapper.writeValueAsString(WorkflowController.getReport()), JSONCompareMode.STRICT);
 		verify(ddotService).parseDdot(any(MultipartFile.class));
 		verify(transformService).transform(anyMap());
+		//verify(legacyCruService).getMonitoringLocation(anyMap(), anyString());
```
cschroed-usgs: Should this be commented out, or actually run? If the former, should this commented-out code be removed?

kmschoep-usgs: good catch!

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
```diff
@@ -59,15 +59,17 @@ public String updateTransaction(Object agencyCode, Object siteNumber, String jso
 		}
 	}
 	
-	public Map<String, Object> getMonitoringLocation(Object agencyCode, Object siteNumber) {
+	public Map<String, Object> getMonitoringLocation(Object agencyCode, Object siteNumber, boolean isAddTransaction) {
```
cschroed-usgs: Are there tests that call this method with both `isAddTransaction=true` and `isAddTransaction=false`?

cschroed-usgs: Self-answer: I only see tests where this method is explicitly called with `isAddTransaction=false`. I can't tell if there are  higher-level tests that would implicitly call this method with `isAddTransaction=true`. Is it worthwhile to add one or more tests where this method is called with `isAddTransaction` explicitly set to `true`?

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
```diff
@@ -59,15 +59,17 @@ public String updateTransaction(Object agencyCode, Object siteNumber, String jso
...
 		Map<String, Object> site = new HashMap<>();
 		ObjectMapper mapper = new ObjectMapper();
 		
 		ResponseEntity<String> cruResp = legacyCruClient.getMonitoringLocation((String)agencyCode, (String)siteNumber);
 		int cruStatus = cruResp.getStatusCodeValue();
 		
 		if (cruStatus == 404) {
-			BaseController.addStepReport(new StepReport(SITE_GET_STEP, cruStatus,  SITE_GET_DOES_NOT_EXIST , agencyCode, siteNumber));
+			if (!isAddTransaction) {
```
zmoore-usgs: Is there ever a case where we'd get a 404 and it is an add transaction? If so is there any sort of step report we'd want to add here (something like "Site not found (Expected)") or perhaps just a log message to let us know what happened?

cschroed-usgs: Yeah, this is tricky. I like the idea of a log message as a compromise for now. Someday I think it would be cool if the workflow could succeed when some steps had non-2xx status codes. If there is time left in the sprint, I would support you pursuing that improvement in a subsequent pull request. What do you think?

kmschoep-usgs: The workflow does currently succeed with non-2xx status codes.  It's just what is displayed to the user leads them to believe it doesn't.  There is another task to completely revamp this output, so I want to keep this scope narrow.

kmschoep-usgs: This isn't an edge case, it's the happy path.

zmoore-usgs: Yeah, I agree the message is certainly confusing when a 404 is a success in one case and a failure in another. Hopefully in the revamp output task we can come up with an output that abstracts away HTTP status codes altogether, and then whomever looks at the HTTP-request level output (assuming we keep both available) will know what the actual statuses mean in the context of each workflow.

cschroed-usgs: Yep. Sorry. I got confused from context-switching. Someday we might want to distinguish between 404's returned due to bad URLs (bad web service URLs in env vars) vs 404's for URLs that do hit the right URL for the web service but which don't return a monitoring location. I am happy merging without that distinction today though.

kmschoep-usgs:
see json output in JIRA task new_site_no_dup_200.json,

cschroed-usgs:
Clever! Nice fix 🌟

kmschoep-usgs:
agreed

PR #61, opened 2018-11-14
Possible fix.

cschroed-usgs:
No idea if this works. Having trouble with local setup.


cschroed-usgs:
I got the gateway running locally via a series of ugly hacks. It appears this fixes the 500 problem.

PR #62, opened 2018-11-15
MLR-359: Export workflow now places transaction file in S3.

kmschoep-usgs:


kmschoep-usgs:
None of the tests caught this issue.

cschroed-usgs:
Ok! Is it worthwhile to add tests to prevent this bug from being reintroduced in the future?

PR #65, opened 2018-11-26
Switched Export Workflow UI to use "Copy" language. Added help tooltip.

cschroed-usgs:


src/main/resources/static/index.html
```diff
@@ -42,13 +42,14 @@ <h3 id="ddot-header" class="mlr-action-header">Upload and Process a Ddot File</h
...
-						<h3 id="export-header" class="mlr-action-header">Export a Location to a New Host</h3>
+						<h3 id="export-header" class="mlr-action-header">Copy a Location to a New Host</h3>
 						<form name="exportForm" id="exportForm">
 							<label for="exportAgency">Agency Code:</label>
 							<input id="exportAgency" type="text" name="exportAgency" class="mlr-text-input"><br>
 							<label for="exportSite">Site Number:</label>
 							<input id="exportSite" type="text" name="exportSite" class="mlr-text-input"><br>
-							<input type="button" id="export-button" value="Export" class="btn btn-xs btn-primary mlr-button-input" onclick="exportLocation()">
+							<button type="submit" id="export-button" class="btn btn-xs btn-primary mlr-button-input export-buttons" onclick="exportLocation()">Copy</button>
+							<span id="export-help" class="btn btn-xs btn-info export-buttons" data-toggle="tooltip" data-placement="right" title="Look up a Monitoring Location in the Registry, copy it to a legacy host."><span class="glyphicon glyphicon-question-sign"></span></span>
```
cschroed-usgs: The `title` attribute's value is displayed in the tooltip. I don't know what is supposed to go here. Could any improvements be made to this user-facing description?

src/main/resources/static/ui.js
```diff
@@ -145,4 +145,20 @@ function exportLocation() {
...
 	return JSON.stringify(response, null, 4).split("\n").join("<br/>").split(" ").join("&nbsp;");
-}
\ No newline at end of file
+}
+
+$(function () {
+	//initialize bootstrap tooltips
+	$('[data-toggle="tooltip"]').tooltip();
+	
+	$('#exportForm').submit(function(e){
```
zmoore-usgs: If we include this do we still need the `onclick` HTML attribute on the button?

src/main/resources/static/index.html
zmoore-usgs: I'd probably ask Jessica or Kevin what they'd prefer to see there. I'm not sure what terminology the LDMs are used to using for this sort of action

src/main/resources/static/ui.js
cschroed-usgs: Nope! Good catch. I removed the `onclick` attribute.

PR #67, opened 2018-11-27
Attach report to email rather than including it in the body

cschroed-usgs:
Updated notification microservice call to use new attachment functionality.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
@@ -28,6 +28,7 @@ public class NotificationService {
 	public static final String NOTIFICATION_STEP = "Notification";
 	public static final String NOTIFICATION_SUCCESSFULL = "Notification sent successfully.";
 	public static final String NOTIFICATION_FAILURE = "Notification failed to send.";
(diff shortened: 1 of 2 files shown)


src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -28,6 +28,7 @@
 	public static final String NOTIFICATION_STEP = "Notification";
 	public static final String NOTIFICATION_SUCCESSFULL = "Notification sent successfully.";
 	public static final String NOTIFICATION_FAILURE = "Notification failed to send.";
+	public static final String ATTACHMENT_FILE_NAME = "mlr.json";
```
zmoore-usgs: Wonder if it would be useful to name this something in relation to the DDot file or agency code and site number? That way when LDMs download the JSON files they know which JSON corresponds to which DDot they submitted if they uploaded a bunch at once. Might be considered out-of-scope for this task, but could be an improvement worth asking Kevin about

zmoore-usgs: Yep, sounds like a good plan to me

cschroed-usgs: 👍

PR #72, opened 2018-12-06
Mlr 331

kmschoep-usgs:


src/main/java/gov/usgs/wma/mlrgateway/StepReport.java
```diff
@@ -8,7 +8,7 @@
 	private String name;
 	private Integer httpStatus;
 	private boolean success;
-	private String details;
+	private Object details;
```
zmoore-usgs: What's the reasoning here? Is this to support JSON from other services like the validator so that it isn't as badly stringified?

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+	@Autowired
+	public UserSummaryReportBuilder(){
+	}
+
+	public UserSummaryReport buildUserSummaryReport(GatewayReport report) {
+		UserSummaryReport userSummaryReport = new UserSummaryReport();
+		List<SiteReport> reportedSites = new ArrayList<>();
+		
+		try {
+		List<SiteReport> successfulSites = report.getSites().stream()
```
zmoore-usgs: Indentation seems a little wonky here. I think the stuff inside the `try{...}` block should be indented one level further?

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+				.collect(Collectors.toList());
+		
+		userSummaryReport.setInputFileName(report.getInputFileName());
+		userSummaryReport.setName(report.getName());
+		userSummaryReport.setReportDateTime(report.getReportDateTime());
+		userSummaryReport.setUserName(report.getUserName());
+		userSummaryReport.setNumberSiteSuccess(successfulSites.size());
+		userSummaryReport.setNumberSiteFailure(failureSites.size());
+		userSummaryReport.setWorkflowSteps(failureWorkflowSteps);
+
```
zmoore-usgs: Extra blank line here

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+				.filter(s -> s.getSteps().stream()
+						.anyMatch(st -> st.getDetails().toString().contains("warning")) || !s.isSuccess())
+				.collect(Collectors.toList());
+		
+		for (SiteReport site : warningFailureSites) {
+			List<StepReport> warningFailureSteps = site.getSteps().stream()
+			.filter(st -> st.getDetails().toString().contains("warning") || !st.isSuccess())
+			.collect(Collectors.toList());
+			// clear out existing steps
+			site.setSteps(null);
```
zmoore-usgs: Is this line necessary since we're using `setSteps` on line 55?

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+			// add just the warning/failure steps
+			site.setSteps(warningFailureSteps);
```
zmoore-usgs: Wondering if we might want some sort of copy constructor or some other deep-copy method for steps? Wondering if changing the site steps here might end up also changing them in the Gateway report since these are the step reports pulled directly from the input GatewayReport and they might be pulled by reference (i.e: Thinking that the SiteReports here are not copies of the reports from the GatewayReport, but actually the same ones thus changing these is equivalent to changing the GatewayReport steps as well)

src/main/java/gov/usgs/wma/mlrgateway/StepReport.java
kmschoep-usgs: Yeah, was hoping that would be the case, but sadly it is not, so I'll probably put it back as String.

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
kmschoep-usgs: Huh, yep.

src/main/java/gov/usgs/wma/mlrgateway/StepReport.java
zmoore-usgs: :maracas-sad:

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
kmschoep-usgs: so `List<SiteReport> warningFailureSites` doesn't create a new list of SiteReports?

zmoore-usgs: I'm not totally sure since I'm not sure how Streams work. Perhaps Streams do make copies, but _think_ what it's doing is essentially `warningFailureSites=report.getSites()` which would just copy the list of references to `warningFailureSites` and then filter out the ones that don't meet the criteria. I think we probably need something like:

https://stackoverflow.com/questions/47833095/java-8-list-copy?noredirect=1&lq=1

The first answer in that thread links to another post that explains how to do a deep-copy using streams

zmoore-usgs: Might not be a bad idea to add a test case for this, if possible. Something like create a GatewayReport with SiteReports and pass it through your UserSummaryReport generator and then verify the user summary report has the updated site report details and the original GatewayReport still has the unmodified details

kmschoep-usgs: I had the same thought this morning!

src/main/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+			List<SiteReport> failureSites = report.getSites().stream()
+					.filter(s -> !s.isSuccess())
+					.collect(Collectors.toList());
+			
+			List<SiteReport> warningFailureSites = report.getSites().stream()
+					.filter(s -> s.getSteps().stream()
+							.anyMatch(st -> st.getDetails().toString().contains("warning")) || !s.isSuccess())
+					.collect(Collectors.toList());
+			
+			reportedSites.addAll(warningFailureSites);
```
kmschoep-usgs: So even this doesn't disassociated those sites with the GatewayReport sites.  Java sure is sticky!

zmoore-usgs: Yeah, I think we'll need something like the second answer here:

https://stackoverflow.com/questions/47833095/java-8-list-copy?noredirect=1&lq=1

zmoore-usgs:
@kmschoep-usgs I created a PR to your branch with my changes for MLR-373.

https://github.com/kmschoep-usgs/MLR-Gateway/pull/1

src/main/java/gov/usgs/wma/mlrgateway/util/ClientErrorParser.java
```diff
@@ -0,0 +1,50 @@
...
+	public String parseClientError(String key, String clientErrorMessage) {
+
+		ObjectMapper mapper = new ObjectMapper();
+		TypeReference<Map<String, List<Object>>> mapType = new TypeReference<Map<String, List<Object>>>() {};
+		Map<String, List<Object>> map = new HashMap<>();
+		String rtn;
+		try {
+			map = mapper.readValue(clientErrorMessage, mapType);
+			rtn = mapper.writeValueAsString(map.get(key).get(0));
+		} catch (JsonGenerationException e) {
```
zmoore-usgs: You can combine multiple exception types into a single catch statement: 

`catch(JsonGenerationException | JsonMappingException | IOException e) { ...`

src/main/java/gov/usgs/wma/mlrgateway/util/ClientErrorParser.java
```diff
@@ -0,0 +1,50 @@
...
+			rtn = clientErrorMessage;
+			e.printStackTrace();
```
zmoore-usgs: Could make use of the configured logger from line 22 here with something like `log.error("An error occurred while trying to parse a client error message: ", e);` This should help print the log with the proper logger level as well.

src/main/java/gov/usgs/wma/mlrgateway/GatewayReport.java
```diff
@@ -24,6 +25,15 @@ public GatewayReport(String name, String inputFileName) {
 		workflowSteps = new ArrayList<>();
 	}
 	
+	public GatewayReport(GatewayReport gatewayReport) {
+        this.name = gatewayReport.name;
```
zmoore-usgs: Some funky indentation here

src/main/java/gov/usgs/wma/mlrgateway/UserSummaryReport.java
```diff
@@ -0,0 +1,30 @@
+package gov.usgs.wma.mlrgateway;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class UserSummaryReport extends GatewayReport {
+
+	private Integer numberSiteSuccess;
```
cschroed-usgs: One oddity of Java is that `Integer` and `int` types can often be used interchangeable thanks to the Autoboxing/Autounboxing feature.

https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

No need to change anything, just thought I would bring it up.

src/main/java/gov/usgs/wma/mlrgateway/util/ClientErrorParser.java
```diff
@@ -0,0 +1,50 @@
...
+		} catch (JsonMappingException e) {
+			rtn = clientErrorMessage;
+			e.printStackTrace();
```
cschroed-usgs: `printStackTrace()` is handy for debugging, but it isn't usually included in production code. Some code quality tools flag it as a thing to remove. Sonar has a rule with some expansion on it [here](https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1148). Could the `printStackTrace()` calls be replaced with a logging statement that logs the exception?

kmschoep-usgs: yep, I've got it logging locally

cschroed-usgs: 🎉

src/main/resources/static/ui.js
```diff
@@ -2,6 +2,7 @@ var reportUrl = null;
 var uploadRequest = false;
 
 function startLoading(headerText) {
+	document
```
cschroed-usgs: Could this line be removed?

src/main/resources/static/ui.js
```diff
@@ -145,7 +146,100 @@ function exportLocation() {
...
+			if (w.name === "Complete Export Workflow"){
+				if (w.success === false) {
+					workflowFailureMsg = "<b>" + w.name + " Failed: " + JSON.parse(w.details).error_message + "</b><br/><br/>";
+				}
+			}
+		});
+		responseStr = responseStr + workflowFailureMsg;
+	} else {
+		if (response.workflowSteps.length > 0){
+			var workflowFailure = response.workflowSteps.filter(function(x){
```
cschroed-usgs: Will this always be just one failure? If there can be more than one, could the variable be named  `workflowFailures` instead?

kmschoep-usgs: Yes, this is to grab any overall workflow failure steps.  There is only one workflow per submission.

src/main/resources/static/ui.js
kmschoep-usgs: apparently so.

src/main/resources/static/ui.js
cschroed-usgs: Got it! Thanks!

src/main/resources/static/ui.js
```diff
@@ -145,7 +146,100 @@ function exportLocation() {
 }
 
 function formatJsonResponse(response) {
-	return JSON.stringify(response, null, 4).split("\n").join("<br/>").split(" ").join("&nbsp;");
+	var responseStr; 
+	var workflowErrorRow = "";
+	var siteErrorRows = "";
+
```
cschroed-usgs: Some people are strict about banning br elements from their codebase on the grounds that it mixes presentation and content. I don't want to draw a hard line about it, but could you try converting some of these br's to spans, divs, or a dl to see if any are easy wins?  More on dl's : https://www.w3schools.com/tags/tag_dl.asp

src/test/java/gov/usgs/wma/mlrgateway/util/UserSummaryReportTest.java
```diff
@@ -0,0 +1,421 @@
...
+		stepReports.addAll(Arrays.asList(siteTransformerSuccess, siteStationNameDuplicateValidationSuccess, siteValidationSuccess));
+		
+		SiteReport siteReport1 = new SiteReport();
+		siteReport1.setSuccess(true);
+		siteReport1.setAgencyCode("USGS");
+		siteReport1.setSiteNumber("12345678");
+		siteReport1.setTransactionType("A");
+		siteReport1.setSteps(stepReports);
+		gatewayReport.addSiteReport(siteReport1);
+
```
cschroed-usgs: Thanks for creating all of these tests! I had a hard time figuring out what was the same about site reports across tests vs. what varied across tests. To better highlight that, and to make things shorter, is there anything common enough between the different test SiteReports that could be put into a method?
```
private SiteReport makeSiteReport1() {
		SiteReport siteReport1 = new SiteReport();
		siteReport1.setSuccess(true);
		siteReport1.setAgencyCode("USGS");
		siteReport1.setSiteNumber("12345678");
		siteReport1.setTransactionType("A");
                return siteReport1;
}
```
Example Usage:
```
	@Test
	public void successWorkflow_goodSite() throws Exception {
		stepReports = new ArrayList<>();
		stepReports.addAll(Arrays.asList(siteTransformerSuccess, siteStationNameDuplicateValidationSuccess, siteValidationSuccess));

		SiteReport siteReport1 = makeSiteReport1();
		siteReport1.setSteps(stepReports);
		gatewayReport.addSiteReport(siteReport1);
...
```

src/main/resources/static/ui.js
cschroed-usgs: 😆

cschroed-usgs:
Congrats on leveling up the code coverage on this app while making pretty substantial changes. I noticed that the StepReport's setter methods aren't hit in any of the tests.

https://coveralls.io/builds/20725120/source?filename=src%2Fmain%2Fjava%2Fgov%2Fusgs%2Fwma%2Fmlrgateway%2FStepReport.java

Since code coverage is so good in general, I'm guessing that the setter's might not be used at all. Would it make sense to go the immutable route with that class? (add the `final` keyword to all properties, remove the setters). There's blogs a-plenty about the benefits of immutability. Here's just one:
https://dzone.com/articles/taming-concurrency-with-immutability

PR #76, opened 2019-02-01
MLR-381: Add user-friendly report output to email body.

kmschoep-usgs:


src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +91,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+	
+	protected String parseSiteErrors(List<SiteReport> siteErrors) {
+		String siteReportRows = "";
+		String errorRow = "";
+		String warningRow = "";
+		String siteAgencyNumber = "";
+		for (SiteReport site : siteErrors) {
+			siteAgencyNumber = site.getAgencyCode().trim() + "-" + site.getSiteNumber().trim();
+			for (StepReport siteStep : site.getSteps()) {
+				errorRow += getDetailSiteMessage(siteAgencyNumber,siteStep, "error");
```
zmoore-usgs: May want to convert these key strings into an ENUM or a boolean if there's only the two possibilities.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +91,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+		String result = "";
+		if (messageType.equals("error") && siteStep.getDetails().toString().contains("error_message")) {
+			try {
+				jsonMap = objectMapper.readValue(siteStep.getDetails(),
+					new TypeReference<Map<String,String>>(){});
+				if (jsonMap.containsKey("error_message")) {
+					result += site + ", " + siteStep.getName() + " Fatal Error: " + jsonMap.get("error_message");
+				}
+			} catch (IOException e) {
+				log.error(NOTIFICATION_STEP + ": error message might be object, trying to parse; " + e.getMessage());
```
zmoore-usgs: May want to log these as warnings or infos since it's not really an error until it fails the very last catch try/catch.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +91,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+		return result;
+	}
+	
+	protected String getDetailSiteMessage(String site, StepReport siteStep, String messageType) {
+		ObjectMapper objectMapper = new ObjectMapper();
+		Map<String, String> jsonMap = new HashMap<>();
+		Map<String, Map<String, String>> jsonObjMap = new HashMap<>();
+		Map<String, Map<String, Map<String, List<String>>>> jsonValMap = new HashMap<>();
+		String result = "";
+		if (messageType.equals("error") && siteStep.getDetails().toString().contains("error_message")) {
```
zmoore-usgs: I'm not really a fan of how free-form we've let our error responses become to where we need so many layers of parsing. I won't advocate for changing this now since it works as-is and we're rapidly nearing the end of this part of this project, but for the next project I think we may want to take a step back and come up with a standard and well-defined error response format that we can make all of our services adhere to. Doing that would let us define an "error response" pojo in the gateway that could be used here to avoid the sprawl of parsing.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +91,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+						for (Map.Entry<String, String> entry : jsonObjMap.get("error_message").entrySet()) {
+							result += site + ", " + siteStep.getName() + " Fatal Error: " + entry.getKey() + " - " + entry.getValue() + "\n";
+						}
+					}
+				} catch (IOException e1) {
+					log.error(NOTIFICATION_STEP + ": error parsing object error message, could be validation message; " + e1.getMessage());
+					try {
+						jsonValMap = objectMapper.readValue(siteStep.getDetails(),
+								new TypeReference<Map<String,Map<String, Map<String, List<String>>>>>(){});
+						if (jsonValMap.containsKey("validator_message")){
```
zmoore-usgs: May want to define these hard-coded key strings as constants.

"validator_message", "warning_message", "error_message", "fatal_error_message", etc.

src/test/java/gov/usgs/wma/mlrgateway/service/NotificationServiceTest.java
```diff
@@ -68,48 +81,355 @@ public void happyPath() throws Exception {
...
 		List<String> recipientList = new ArrayList<>();
 		recipientList.add("test");
-		service.sendNotification(recipientList, "test", "test", "test", BaseController.getReport());
+		service.sendNotification(recipientList, "test", "test", "test", userSummaryReport);
 	}
 
 	@Test
 	public void buildMessageBodyTest() throws Exception {
-		GatewayReport report = new GatewayReport("report-name", "report-file-name");
+		userSummaryReport.setReportDateTime("report-date-time");
```
zmoore-usgs: Appreciate the thorough test coverage!

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -64,7 +69,7 @@ public static void remove() {
 	protected void notificationStep(String subject, String attachmentFileName) {
 		List<String> notificationRecipientList;
 		String userName = "Unknown";
-		
+		userSummaryReport = userSummaryReportBuilder.buildUserSummaryReport(getReport());
```
zmoore-usgs: Not sure if it matters or not - Should the `getReport()` call happen within the try/catch block as it was before or was it intentionally moved outside of it or does it not matter?"

Not sure if that block was doing or should be doing any error handling for any exceptions that might be thrown during the report generation.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
kmschoep-usgs: good point

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
kmschoep-usgs: I agree, and sad part is that I already did spend a good bit of time getting them standardized even this much.

cschroed-usgs:
@kmschoep-usgs , thanks for the much-needed work to improve user-facing output and thanks for your efforts to standardize messages! The struggle is real. It sounds like we all broad-strokes agree that we should improve things further in the future. To keep this pull request to a reasonable scope and to ensure that we follow up on this, I encourage you both to brainstorm solutions in MLR-393.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +91,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+				new TypeReference<Map<String,String>>(){});
+			result = jsonMap.get("error_message");
+		} catch (IOException e) {
+			log.error(NOTIFICATION_STEP + ": error parsing error message; " + e.getMessage());
+				result = detailMessage;
+		}
+		return result;
+	}
+	
+	protected String getDetailSiteMessage(String site, StepReport siteStep, String messageType) {
```
cschroed-usgs: Could this method be broken up into into a few smaller descriptively-named helper methods? A few method names that come to mind:
handleDetailSiteErrorMessage
handleDetailSiteWarningMessage

kmschoep-usgs: It could be.  I was hoping to be able to reuse some code, but that didn't happen.

src/test/java/gov/usgs/wma/mlrgateway/service/NotificationServiceTest.java
cschroed-usgs: I agree! Thank you for dedicating time to these extensive tests!

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -84,4 +96,139 @@ protected String buildAttachmentName(String attachmentFileName) {
...
+			}
+		}
+		siteReportRows += errorRow + warningRow ;
+		return siteReportRows;
+		
+	};
+	
+	protected String getDetailErrorMessage(String detailMessage) {
+		ObjectMapper objectMapper = new ObjectMapper();
+		Map<String, String> jsonMap = new HashMap<>();
```
isuftin: Because this variable is not used outside of the try/catch block, you might want to declare it within that block to narrow the scope where it's used. 

You can also move the objectMapper variable within the try block and have it be `new ObjectMapper().readValue(detailMessage, new TypeReference<Map<String,String>>(){});`

Also, depending on how you view returns from functions, you could return on line 167 and on line 170 and remove line 172.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
```diff
@@ -32,13 +38,18 @@
 	public static final String NOTIFICATION_FAILURE = "{\"error_message\": \"Notification failed to send.\"}";
 	public static Temporal reportDateTime;
 	public static final String ATTACHMENT_FILE_NAME = "mlr-%NAME%-report.json";
+	public enum MessageType {error, warning};
```
cschroed-usgs: Could these enum values please be upper-cased to make them a bit more idiomatically Java-ish?
"Because they are constants, the names of an enum type's fields are in uppercase letters."
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

isuftin:
I am seeing a lot of string concatenation in this PR. If you're doing a lot of String concatenation in Java, you may consider using StringBuilder instead.

http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html

src/test/java/gov/usgs/wma/mlrgateway/service/NotificationServiceTest.java
```diff
@@ -68,48 +81,355 @@ public void happyPath() throws Exception {
...
+		sites.add(siteReport1);
+		sites.add(siteReport2);
+		report.setName("Validate D dot File");
+		report.setUserName("report-user");
+		report.setInputFileName("report-file-name");
+		report.setReportDateTime("report-date-time");
+		report.setWorkflowSteps(workflowSteps);
+		report.setSites(sites);
+		report.setNumberSiteSuccess(0);
+		report.setNumberSiteFailure(1);
```
cschroed-usgs: Is it worthwhile to scoot any of the common report setup code into a method? I know that the returns diminish pretty quickly when investing in the deduplication of testing code, but I thought I would ask.

cschroed-usgs:
> I am seeing a lot of string concatenation in this PR. If you're doing a lot of String concatenation in Java, you may consider using StringBuilder instead.
> 
> http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html

Interesting article. I thought that the JVM effectively did that for you these days, but it looks like that isn't the case or at least it wasn't in 2015. The plot shows nontrivial differences in performance starting around 10k iterations. @kmschoep-usgs are we concatenating enough strings where this would start to matter?

Another approach that could be taken for reasons of performance or maintainability would be to use basic string concatenation to build each message, append each error message to an `ArrayList<String> messages`, and then `return String.join("\n", messages)`. It would prevent the "I forgot to add that freaking newline/comma/delimiter char" kind of problem that I have bashed my head against in the past.

isuftin:
Re; String concat, I tend to just always use recommended best practice and use StringBuilder. But both Java String concatenation and using StringBuilder does make things hard to read when reading code, and I hate that.

I much prefer using something like Java's MessageFormat which is somewhat like Python's format() function

http://www.java2s.com/Code/JavaAPI/java.text/MessageFormatformatStringpatternObjectarguments.htm

Though I don't know what it does behind the scenes, but at least it's a lot more readable.

src/main/java/gov/usgs/wma/mlrgateway/service/NotificationService.java
cschroed-usgs: I sympathize! In this case the helper methods wouldn't help re-use, but they could increase adherence to the [Single Level of Abstraction Principle](http://principles-wiki.net/principles:single_level_of_abstraction) and thereby improve readability.

Another related thing to consider...
Though reasonable people may disagree, many Java programmers prefer to avoid nested try/catches. The following links show some of the debate and the solutions. Most advocate for refactoring into smaller methods.

https://softwareengineering.stackexchange.com/questions/237444/alternatives-to-nested-try-catches-for-fallbacks
https://softwareengineering.stackexchange.com/questions/118788/is-using-nested-try-catch-blocks-an-anti-pattern
https://www.veracode.com/blog/secure-development/avoiding-nested-try-catch-java

isuftin:
Putting off?

kmschoep-usgs:
> Putting off?

for now.  Need to ponder which method I prefer.

isuftin:
👍

PR #81, opened 2019-02-19
add utility that gets the username and email of logged-in user.

kmschoep-usgs:
first attempt.  unsure of its threadsafeness.


zmoore-usgs:
Looks like there are still a few test failures, but once those are fixed I think this should be good 👍

src/main/java/gov/usgs/wma/mlrgateway/util/GetSecurityContext.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Map;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.HashMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Service
+public class GetSecurityContext {
```
cschroed-usgs: For better [or for worse](https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html), class names in Java are typically nouns. Could this class' name be changed to a noun? `SecurityContext`? `MlrSecurityContext`?

src/main/java/gov/usgs/wma/mlrgateway/util/GetSecurityContext.java
```diff
@@ -0,0 +1,55 @@
...
+	
+	private Logger log = LoggerFactory.getLogger(GetSecurityContext.class);
+	
+	@Autowired
+	public GetSecurityContext(){
+	}
+
+	private Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
zmoore-usgs: Looking at this more, I'm not actually sure this will work. This bit here will run when an instance of the GetSecurityContext class is created. Because this is annotated with `@Service` the instance is created as a Bean when the app is started. Thus the only context this will get is right when the app starts. I think this method (`SecurityContextHolder.getContext().getAuthentication()`) needs to be run within the thread that the user is connected on, and running here that will not be the case.

You could probably add `getUserName()` and `getUserEmail()` as methods within the Base Controller and have each of those methods actually run `SecurityContextHolder.getContext().getAuthentication()`

cschroed-usgs: Where possible (sometimes it's not), it is preferable to retrieve an object's dependencies via dependency injection. These two blog posts show how to get `Authentication` objects via dependency injection instead of via static method calls.
https://www.baeldung.com/get-user-in-spring-security
https://dzone.com/articles/how-to-get-current-logged-in-username-in-spring-se

Both of those blog posts are written in the context of `@Controller`-annotated classes though. Is it possible to to inject the `Authentication` object inside a `@Service`-annotated class like this one?

src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
```diff
@@ -34,7 +36,8 @@
 	private UserSummaryReportBuilder userSummaryReportbuilder;
 	public static final String COMPLETE_WORKFLOW_SUBJECT = "Submitted Ddot Transaction";
 	public static final String VALIDATE_DDOT_WORKFLOW_SUBJECT = "Submitted Ddot Validation";
-
+	private GetSecurityContext getSecurityContext = new GetSecurityContext();
```
cschroed-usgs: Is it possible/relevant to remove these classes' manual instantiations of `GetSecurityContext`  and instead inject it? I'm generally a fan of constructor injection. If that's your idea of a good time too, you could add a `GetSecurityContext` parameter to your `@Autowired`-annotated constructors. You might want to make GetSecurityContext a "request"-scoped bean to ensure the object only contains details about the current HTTP request.
Ex:
```
...
@Service
@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
public class GetSecuirtyContext...
```

src/main/java/gov/usgs/wma/mlrgateway/util/GetSecurityContext.java
kmschoep-usgs: OK, that's exactly the kind of confirmation I needed.

cschroed-usgs: Good point Zack. The default scope for `@Service` classes is `singleton`, so the current approach probably wouldn't work. I proposed a way that would (probably) work in the other comments that involves switching the bean's scope to `request`. The `BaseController` way would also work. I would prefer the `BaseController` got a hold of its `Authentication` object via injection rather than static methods.

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -68,6 +66,28 @@ public static void remove() {
 		gatewayReport.remove();
 	}
 	
+	public String getUserName() {
```
cschroed-usgs: These methods look pretty good in the `BaseController`! Since they're in a `@Controller` now, the blog posts on injecting  `Authentication` should (🤞)  be directly applicable.

src/main/java/gov/usgs/wma/mlrgateway/config/WebMvcConfig.java
```diff
@@ -31,4 +32,9 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
 				.resourceChain(true)
 				.addResolver(new WebJarsResourceResolver());
 	}
+	
+	@Bean
+	public Clock clock() {
+	 return Clock.systemDefaultZone();
```
zmoore-usgs: Nit-pick: Looks like the indentation got a little wonky here

cschroed-usgs:
@zmoore-usgs That would work. According to the pages I linked to it should be possible to forego adding Authentication objects as parameters to the methods, and instead inject an Authentication object into the controller. So something like
```
public abstract class BaseController {

protected Authentication authentication;

 public BaseController(Authentication authentication){
   this.authentication = authentication;
}

//getUsername and getEmail methods here

}

@Controller 
public class MyController extends BaseController{


@Autowired
public MyController(Authentication authentication) {
   super(authentication);
}

@PostMapping("/ddots")
	public UserSummaryReport legacyWorkflow(@RequestPart MultipartFile file, HttpServletResponse response) {
		setReport(new GatewayReport(LegacyWorkflowService.COMPLETE_WORKFLOW
				,file.getOriginalFilename()
				,getUserName()

}

}
```

cschroed-usgs:
The `Clock` changes look great @kmschoep-usgs !

cschroed-usgs:
Never mind! The approach Zack suggested is consistent with blog-town, so that is probably the better approach.

src/main/java/gov/usgs/wma/mlrgateway/controller/ExportWorkflowController.java
```diff
@@ -28,15 +29,17 @@
...
 public class ExportWorkflowController extends BaseController {
 	private ExportWorkflowService export;
+	private Authentication authentication;
 	public static final String COMPLETE_WORKFLOW = "Complete Export Workflow";
 	public static final String EXPORT_WORKFLOW_SUBJECT = "Transaction File Generation for Requested Location";
 	private final Clock clock;
 	
 	@Autowired
-	public ExportWorkflowController(ExportWorkflowService export, NotificationService notificationService, Clock clock) {
+	public ExportWorkflowController(ExportWorkflowService export, NotificationService notificationService, Clock clock, Authentication authentication) {
```
zmoore-usgs: Sorry, my comment wasn't super clear. I think `Authentication` shouldn't be a member variable of this class and rather than having it as an argument here it should be an argument on the actual `exportWorkflow` and `legacy...Workflow` rest functions. Then I think it will be injected properly

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -65,10 +69,30 @@ public static void addWorkflowStepReport(StepReport stepReport) {
...
+			log.warn("No Authentication present in the Web Security Context when getting user name.");
+		}
+		return userName;
+	}
 	
-	protected void notificationStep(String subject, String attachmentFileName) {
+	public String getUserEmail(Authentication authentication) {
+		String userEmail = "";
+		if(authentication != null){
+			Map<String, Serializable> oauthExtensions = ((OAuth2Authentication) authentication).getOAuth2Request().getExtensions();
```
kmschoep-usgs: My tests keep bombing on this with ClassCastException, which is understandable, but there doesn't seem to be a way to mock this event.  Am I in interface/implementation land now?  Seems heavy-handed, but....?

zmoore-usgs: What you might be able to do is in the test change the `@MobckBean Authentication` to an `@MockBean OAuth2Authentication`

kmschoep-usgs: now getting NPE when trying to execute this line.  Like it's not really being mocked?

zmoore-usgs: I'm not 100% sure, but I think when you mock you can't mock nested calls like `getOAuth2Request().getExtensions()` directly. I think you might need to create a MockBean of whatever type `getOAuth2Request()` returns. Then mock the `getOAuth2Request()` method to return that mock bean, and mock the `getExtensions()` method of that mock bean to return the `testEmail` map

kmschoep-usgs: So, just tried injecting OAuth2Authentication instead of Authentication everywhere, seems to work locally, trying the tests now...

cschroed-usgs:
👍 :shipit:

PR #82, opened 2019-02-25
updated date parsing to display in browser's time zone.  Verbose, but

kmschoep-usgs:
might be less confusing to user.

### DIFF
--- a/src/main/resources/static/ui.js
+++ b/src/main/resources/static/ui.js
@@ -154,11 +154,12 @@ function formatJsonResponse(response) {
 	var responseStr; 
 	var workflowErrorRow = "";
 	var siteErrorRows = "";
+	var reportDate = new Date(response.reportDateTime);
 	
 	responseStr =  "<ul>";
 	responseStr = "<li>MLR Workflow:&nbsp;&nbsp;" + response.name + "</li>";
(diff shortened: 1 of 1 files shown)


src/main/resources/static/ui.js
```diff
@@ -154,11 +154,12 @@ function formatJsonResponse(response) {
 	var responseStr; 
 	var workflowErrorRow = "";
 	var siteErrorRows = "";
+	var reportDate = new Date(response.reportDateTime);
```
cschroed-usgs: According to MDN, "parsing of date strings with the Date constructor (and Date.parse, they are equivalent) is strongly discouraged due to browser differences and inconsistencies. "
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#Syntax

Unfortunately, MDN does not mention a convenient alternative using native JavaScript. In the past I solved the problem using 3rd-party libraries like SugarJS and Moment.js. Browsers and libraries have both advanced since then. I defer to folks with more contemporary front-end chops on what a better solution would be today.

kmschoep-usgs: Given that this is literally the only place that dates are parsed in the javascript I'm hesitant to bring in a 3rd part library for it.  It's basically a convenience for users to have the time converted to their local time zone instead of UTC.

PR #85, opened 2019-04-01
merge docker and service repos

kmschoep-usgs:


docker/certificates/import_certs/README.md
```diff
@@ -0,0 +1,3 @@
+In this folder you should place any certificates that this serivce will need to trust. For the Notification service that includes certificates for:
```
mhines-usgs: copy pasta - notification service

kmschoep-usgs: gah..

docker/config.env
```diff
@@ -0,0 +1,30 @@
+serverContextPath=/
```
mhines-usgs: i think we can ditch all the redundant configs that come from docker-wma-spring-boot-base from here (as well as for the notification service if I missed that before)

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+#download all maven dependencies (this will only re-run if the pom has changed)
+RUN mvn -B dependency:go-offline
+
+COPY src /build/src
+ARG BUILD_COMMAND="mvn -B clean package"
+RUN ${BUILD_COMMAND}
+
+FROM usgswma/wma-spring-boot-base:8-jre-slim-0.0.4
+
+ENV requireSsl=true
```
mhines-usgs: i think these can be pared down to just the ones that aren't redundant to the docker-wma-spring-boot-base

docker/config.env
zmoore-usgs: Yeah, anything in here that's defined with the same value in the Dockerfile for this app or in the base Dockerfile https://github.com/USGS-CIDA/docker-wma-spring-boot-base/blob/master/8/Dockerfile
can be removed. If you're overriding the value defined in either the Dockerfile for this app or the base Dockerfile then you'd leave it here.

kmschoep-usgs: I was wondering about this one, thanks.

Dockerfile
zmoore-usgs: Like Megan mentioned in the config.env file, anything in here that's defined with the exact same name and value as something in the base image Dockerfile can be removed (such as `requireSsl=true`)

https://github.com/USGS-CIDA/docker-wma-spring-boot-base/blob/0.0.4/8/Dockerfile#L37

README.md
```diff
@@ -70,9 +70,21 @@ The related environment variables are listed below:
 * mlrgateway_legacyValidatorServers - comma separated list of url(s) for the Legacy Validator Microservice
 * mlrgateway_legacyCruServers - comma separated list of url(s) for the Legacy CRU Microservice
 * mlrgateway_fileExportServers - comma separated list of url(s) for the File Export Microservice
-* mlrgateway_notificationServers - comma separated list of url(s) for the Notification Microservice
+* mlrgateway_gatewayServers - comma separated list of url(s) for the gateway Microservice
```
zmoore-usgs: I think this still needs to stay as `notificationServers`. This is explaining how to setup the list of notification service URLs that the Gateway should forward to

.travis.yml
```diff
@@ -1,4 +1,38 @@
...
+---
+  sudo: required
 
-script:
-  - mvn clean verify jacoco:report coveralls:report
\ No newline at end of file
+  env:
+    global:
+      - DOCKER_COMPOSE_VERSION=1.22.0
+      - DOCKER_IMAGE=mlr-gateway-service
```
zmoore-usgs: Might change this to just `mlr-gateway` in here and the compose file (i.e: mlr-gateway, mlr-gateway-local-dev, etc.). We've been dropping the `-service` suffix from most of these services since we don't use it when we actually build the images for CHS.

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+  secrets:
+    tomcat_ssl_crt:
+      file: ./docker/certificates/tomcat-wildcard.crt
+    tomcat_ssl_key:
+      file: ./docker/certificates/tomcat-wildcard.key
+
+  services:
+    mlr-gateway-service:
+      build: .
+      image: mlr-gateway-service
```
zmoore-usgs: Like I mentioned in my other comment, I think removing the `-service` suffix from the gateway is reasonable because we don't use it in CHS

docker/config.env
```diff
@@ -0,0 +1,29 @@
+mlrgateway_springFrameworkLogLevel=INFO
```
zmoore-usgs: The one other thing that it might be worth adding here would be the server lists that mlr-local-dev uses which would make it easier for this service to just be built and directly connected to mlr-local-dev services without other changes: 

https://github.com/USGS-CIDA/mlr-local-dev/blob/master/docker-reference/configuration/mlr-gateway/config.env#L9

I did a similar thing with the aqcu-gateway:

https://github.com/USGS-CIDA/aqcu-gateway/blob/master/docker/config.env#L12

PR #89, opened 2019-08-30
Don't Merge: Demonstration of Code Quality Metrics

cschroed-usgs:
This is just a demo for conversation purposes. This build won't succeed until the parent POM has been deployed to a maven repository.


pom.xml
```diff
@@ -176,24 +185,16 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>pl.project13.maven</groupId>
+                <artifactId>git-commit-id-plugin</artifactId>
+                <version>2.2.3</version>
+            </plugin>
```
zmoore-usgs: Was this just added for additional information or is it required to be added for the parent stuff to work?

cschroed-usgs: Huh. Not sure why I added that. Thanks. Good catch!

PR #90, opened 2019-12-16
Adopted maven code quality parent pom.

cschroed-usgs:


Dockerfile
```diff
@@ -5,14 +5,36 @@ FROM maven@sha256:b37da91062d450f3c11c619187f0207bbb497fc89d265a46bbc6dc5f17c02a
...
 
+# copy git history into build image so that sonar can report trends over time
+COPY .git /build
+
 COPY src /build/src
-ARG BUILD_COMMAND="mvn -B clean package"
+
+# the -D option supresses INFO-level logs about dependency downloads. This enables the build to finish within Travis' log length limit.
+# The -P option skips the dependency security check in favor of build stability -- the official NVD server is rate-limited, and external builds lack access to our internal NVD mirror
+ARG BUILD_COMMAND="mvn -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn clean verify jacoco:report -P=\!dependency-security-check"
```
cschroed-usgs: It's kind of awkward to put this travis-specific command here. I can update Jenkins to pass a different `--build-arg BUILD_COMMAND` to it's `docker` command. An alternative is to somehow parameterize the docker-compose run and pass the build command from the `.travis.yml `, but I wasn't sure that was worth it. What do you think?

zmoore-usgs: I don't have a problem with having this here

PR #91, opened 2020-03-02
Made group comparison case-sensitive.

cschroed-usgs:


src/test/java/gov/usgs/wma/mlrgateway/PermissionEvaluatorImplTest.java
```diff
@@ -13,14 +13,14 @@
 import org.springframework.test.util.ReflectionTestUtils;
 
 @RunWith(SpringRunner.class)
-public class PermissionEvaluatorTest {
+public class PermissionEvaluatorImplTest {
```
cschroed-usgs: Renamed to match convention that most IDEs use. Now we can flip between test an implementation classes easy-breezy.

rowle1tj:
👍

PR #97, opened 2020-03-09
Switched from JDBC sessions to JWT

cschroed-usgs:


zmoore-usgs:
Woah, well that turned out to be a lot simpler than I expected. Kind of cool that it's 100% removed code 😅

One thing I'm not sure about - Do you think this will still work as-is when deployed to Test or will we need to update the UI first to save the JWT to some sort of storage/cookie and pass it in with every request?

cschroed-usgs:
Hmmm. Good question. It "just worked" when I ran it via mlr-local-dev with a gateway image based off of this branch. I will double-check tomorrow in case it was a stale image or similar.

cschroed-usgs:
It worked again locally after killing any image caches. 🤷‍♂️ Let's try it out on the test tier, and revert if things break. I wish there were automated ways to test this, but I think it makes sense to delay adding tests until we switch to KeyCloak.

PR #99, opened 2020-03-11
MLR-454:  New services to parse ddot file

kmschoep-usgs:
and pull out distinct district codes.  Alert pops up in UI if more than
one distinct district code in response.


src/main/resources/static/ui.js
```diff
@@ -146,6 +154,19 @@ function uploadDdot() {
 	postDdot("workflows/ddots", "Ddot Validate and Update Response", generalSuccess, generalError);
 }
 
+function preVerification(response) {
+	stopLoading($('.mlr-response-header').text());
+	var districtCodes = response.districtCodes;
+	if (districtCodes.length > 1) {
+		alert(multipleDistrictCodeMsg);
+	}
+	if (action === "validate"){
```
zmoore-usgs: Rather than using string comparison here maybe make the button action a boolean like `doUpload` and if it's false then it's a validate and otherwise it's a full upload?

src/main/resources/static/ui.js
```diff
@@ -146,6 +154,19 @@ function uploadDdot() {
 	postDdot("workflows/ddots", "Ddot Validate and Update Response", generalSuccess, generalError);
 }
 
+function preVerification(response) {
+	stopLoading($('.mlr-response-header').text());
+	var districtCodes = response.districtCodes;
+	if (districtCodes.length > 1) {
+		alert(multipleDistrictCodeMsg);
```
zmoore-usgs: Not sure if it's part of the requirements or not, but could consider making a choice pop-up using `confirm()` rather than `alert()` that lets the user cancel if they want to?

I.E: https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_confirm2

src/main/java/gov/usgs/wma/mlrgateway/service/PreVerificationService.java
```diff
@@ -0,0 +1,53 @@
...
+@Service
+public class PreVerificationService {
+
+	private DdotClient ddotClient;
+	private Logger log = LoggerFactory.getLogger(PreVerificationService.class);
+
+	protected static final String STEP_NAME = "Ingest D dot File";
+	protected static final String SUCCESS_MESSAGE = "D dot file parsed successfully.";
+	protected static final String INTERNAL_ERROR_MESSAGE = "{\"error_message\": \"Unable to read ingestor output.\"}";
+	protected static final String WRONG_FORMAT_MESSAGE = "{\"error_message\":\"Only accepting files with one transaction at this time.\",\"data\":%ddotResponse%}";
```
zmoore-usgs: Looks like this and `SUCCESS_MESSAGE` aren't used anywhere? Those are used in the other services when the StepReport is being generated, but since this service doesn't get used by the step report workflows I think those can be removed.

src/main/resources/static/ui.js
kmschoep-usgs: sure, can do that.

src/main/java/gov/usgs/wma/mlrgateway/service/PreVerificationService.java
kmschoep-usgs: oops, yep, can pull all that out.

src/main/java/gov/usgs/wma/mlrgateway/ParsedDistrictCodes.java
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	public List<String> getDistrictCodes() {
+		return districtCodes;
+	}
+
+	public void setDistrictCodes(List<String> districtCodes) {
+		this.districtCodes = districtCodes;
+	}
+
```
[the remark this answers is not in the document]
kmschoep-usgs: yeah, this whole file is going away

src/main/java/gov/usgs/wma/mlrgateway/controller/UtilController.java
```diff
@@ -38,5 +57,40 @@ public String getToken() {
...
+	
+	@ApiOperation(value="Parse ddot file and return the list of district codes")
+	@ApiResponses(value={@ApiResponse(code=200, message="OK"),
+			@ApiResponse(code=400, message="Bad Request"),
+			@ApiResponse(code=401, message="Unauthorized"),
+			@ApiResponse(code=403, message="Forbidden")})
+	@PostMapping("/parse")
+	public ParsedDistrictCodes parseWorkflow(@RequestPart MultipartFile file, HttpServletResponse response) throws IOException {
+		List<String> districtCodes = new ArrayList<>();
+		ParsedDistrictCodes parsedDistrictCodes = new ParsedDistrictCodes();
```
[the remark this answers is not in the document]
kmschoep-usgs: Doing that in part 2, when I add transaction types.

src/main/java/gov/usgs/wma/mlrgateway/service/PreVerificationService.java
```diff
@@ -0,0 +1,53 @@
...
+
+	public List<Map<String, Object>> parseDdot(MultipartFile file) throws HystrixBadRequestException {
+		List<Map<String, Object>> ddots = null;
+		ObjectMapper mapper = new ObjectMapper();
+		TypeReference<List<Map<String, Object>>> mapType = new TypeReference<List<Map<String, Object>>>() {};
+		
+		try {
+			String ddotResponse = ddotClient.ingestDdot(file).getBody();
+			ddots = mapper.readValue(ddotResponse, mapType);
+		} catch (Exception e) {
```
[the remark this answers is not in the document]
zmoore-usgs: Actually never mind I think this makes sense looking at it again. In the other services that's mostly done to add proper output to the step report which we aren't doing here.

PR #100, opened 2020-03-11
don't show popup alert and run validate/update as usual for ddots with a

kmschoep-usgs:
single district code.


zmoore-usgs:
:+1:

PR #101, opened 2020-03-12
Add popup if there is an ADD transaction in the submitted ddot file.

kmschoep-usgs:


kmschoep-usgs:
Right now it's two separate popups.  Don't know if this is preferable to having both messages (if applicable) in one popup.  Something for POST to decide.

src/main/resources/static/ui.js
```diff
@@ -157,12 +158,14 @@ function uploadDdot() {
 function preVerification(response) {
 	stopLoading($('.mlr-response-header').text());
 	var districtCodes = response.districtCodes;
-	var confirmProceed;
+	var transactionTypes = response.transactionTypes;
```
zmoore-usgs: Not a big deal and I don't have a big opinion either way, but I think you could forego the intermittent variables and just do `if(response.districtCodes.length > 1)` and `if(response.transactionTypes.includes('A'))`

PR #102, opened 2020-03-18
Upgraded to Spring Boot 2.x. Did not update to latest Spring Security…

zmoore-usgs:
…. Did not update to JUnit5. Minor cleanup.

Skipped updates to Spring Security and JUnit 5 because while WaterAuth can handle _Resource Server_ clients on Spring Security 5.1, it cannot handle _Authorization Server_ clients because currently the Spring Security 5.1 Authorization Server implementation does not support JWTs as a token format from an IDP Server, which Water Auth serves as in this case.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,29 +10,15 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <additionalparam>-Xdoclint:none</additionalparam>
-        <springfox.version>2.7.0</springfox.version>
+        <springfox.version>2.8.0</springfox.version>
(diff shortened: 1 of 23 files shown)


pom.xml
```diff
@@ -47,41 +33,75 @@
...
             </dependency>
             <dependency>
                 <groupId>org.springframework.cloud</groupId>
                 <artifactId>spring-cloud-dependencies</artifactId>
-                <version>Dalston.SR4</version>
+                <version>Greenwich.SR5</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+
```
cschroed-usgs: I like documenting the reason for the version bumps. Hopefully when Spring updates the versions themselves we can use comments like this to decide to remove these version customizations.

PR #107, opened 2020-03-25
MLR-451: Update Primary Key Update workflow

kmschoep-usgs:
The file export part might change based on how the distributed hosts handle updates.


pom.xml
```diff
@@ -25,7 +25,7 @@
         <repository>
             <id>wma-maven-centralized</id>
             <name>Water Mission Area Snapshots and Releases</name>
-            <url>https://cida.usgs.gov/artifactory/wma-maven-centralized</url>
+            <url>https://artifactory.wma.usgs.gov/artifactory/maven-central</url>
```
zmoore-usgs: I think this should be `wma-maven` instead of `maven-central`

`wma-maven` is for our artifacts, where `maven-central` I believe is just a mirror of actual maven central and won't have any of our stuff

src/main/java/gov/usgs/wma/mlrgateway/workflow/UpdatePrimaryKeyWorkflowService.java
```diff
@@ -0,0 +1,140 @@
...
+			//1. Get old monitoring location
+			LOG.trace("Get old monitoring location");
+			oldMonitoringLocation = legacyCruService.getMonitoringLocation(oldAgencyCode, oldSiteNumber, false, oldSiteReport);
+			
+			if (!oldMonitoringLocation.isEmpty()) {
+				//2. Set new monitoring location
+				LOG.trace("Set new monitoring location");
+				newMonitoringLocation = oldMonitoringLocation;
+				
+				//3.update new monitoring location with new AgencyCode and/or SiteNumber
```
zmoore-usgs: Does this go through the validator at all? We might want to validate `monitoringLocation` once it has its new AC/SN to make sure it won't be creating a duplicate with a AC/SN combo that already exists? Though maybe that bit is handled inherently by the CRU, I can't remember 😅

src/main/java/gov/usgs/wma/mlrgateway/workflow/UpdatePrimaryKeyWorkflowService.java
```diff
@@ -0,0 +1,140 @@
...
+				newMonitoringLocation.replace(AGENCY_CODE, oldMonitoringLocation.get(AGENCY_CODE));
+				newMonitoringLocation.replace(SITE_NUMBER, oldMonitoringLocation.get(SITE_NUMBER));
+				
+				//4.replace station name and site web ready code for old site
+				oldMonitoringLocation.replace(SITE_WEB_READY_CODE, "L");
+				oldMonitoringLocation.replace(STATION_NAME, "DEPRECATED SITE: superceded by " + newAgencyCode.trim() + "-" + newSiteNumber.trim() + ".");
```
zmoore-usgs: Similar to my last comment, should the new name / new stationIx be validated to ensure it's unique? I assume it would be unless somehow multiple sites are deprecated to the same new site (which shouldn't be possible given the unique AC/SN cosntraints so maybe this is a non-issue)

src/main/java/gov/usgs/wma/mlrgateway/workflow/UpdatePrimaryKeyWorkflowService.java
kmschoep-usgs: Ah, good point!  Yes, it should go through the validator.

pom.xml
kmschoep-usgs: hm, OK.  I have it set this way for decap as well, and it seems to work.  I'll update it though.

src/main/java/gov/usgs/wma/mlrgateway/workflow/UpdatePrimaryKeyWorkflowService.java
kmschoep-usgs: probably safest to send them all through the validator.

pom.xml
zmoore-usgs: If it works I suppose it doesn't need to be updated then, I was just thinking that the code quality POM wouldn't be in `https://artifactory.wma.usgs.gov/artifactory/maven-central` and actually it doesn't seem to be in `wma-maven`. In my PR here I changed this to the EROS artifactory mirror and got it to work: https://github.com/USGS-CIDA/MLR-Gateway/pull/108/files#diff-600376dffeb79835ede4a0b285078036R28

zmoore-usgs:
Looks like a slight conflict with the POM.xml, otherwise I think this should be good!

PR #111, opened 2020-04-01
Add login to util controller for new UI

dpattermann-usgs:


src/main/java/gov/usgs/wma/mlrgateway/controller/UtilController.java
```diff
@@ -56,6 +56,26 @@ public String getToken() {
...
+
+	@ApiOperation(value="Return the user to the new UI, logged in.")
+	@ApiResponses(value={@ApiResponse(code=200, message="OK"),
+			@ApiResponse(code=400, message="Bad Request"),
+			@ApiResponse(code=401, message="Unauthorized"),
+			@ApiResponse(code=403, message="Forbidden"),
+			@ApiResponse(code=404, message="Not Found")})
+	@GetMapping("/login")
+	public void login(HttpServletResponse response) {
+		if(environmentTier != null && !environmentTier.equals(ConfigurationValues.ENVIRONMENT_PRODUCTION)) {
```
zmoore-usgs, line 68: I think we'd still want this to work in Production as well. We disabled the token endpoint for production back when we first created this because we were worried it could be a security vulnerability to let users in Production see their token (I think that could be considered a moot point now and we could probably remove this check from that endpoint as well), but I think we still want this login process to work in prod.

dpattermann-usgs: Yeah, I wasn't sure why we were not doing this on Production so I just followed suit. I can remove this. Should I also remove the environmentTier != null?

zmoore-usgs: Yeah, I'd say remove both

src/test/java/gov/usgs/wma/mlrgateway/controller/UtilControllerTest.java
```diff
@@ -1,67 +1,67 @@
-package gov.usgs.wma.mlrgateway.controller;
+// package gov.usgs.wma.mlrgateway.controller;
```
kmschoep-usgs, line 1: Are these commented-out lines still needed or is this a WIP?

dpattermann-usgs: Good point. I'll try running these tests now that we've gotten this figured out. We did this when I was working directly with Zack to figure out what was going on.

zmoore-usgs:
Assuming the public URL for CHS Artifactory is working again (for maven at least) I think we should try changing the `repository` URL(s) in the pom.xml to `artifactory.wma.usgs.gov` from `artifactory.wma.chs.usgs.gov` so that Travis can complete

PR #113, opened 2020-04-17
Change Primary Key Workflow

kmschoep-usgs:
Now does a true update (PUT) of the location record instead of a
create/modify of a deprecated site.

TODO:  Will need to call a new file export service when new transaction file format is determined.


src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
 		}
 	}
 	
+	public void updatePrimaryKeyWorkflow(String oldAgencyCode, String oldSiteNumber, String newAgencyCode, String newSiteNumber) throws HystrixBadRequestException {
```
dpattermann-usgs, line 160: This method is difficult to parse for me, but that may be just because I'm not plugged into this project super well.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
...
+		String json;
+		Map<String, Object> monitoringLocation = new HashMap<>();
+		Map<String, Object> updatedMonitoringLocation = new HashMap<>();
+		SiteReport siteReport = new SiteReport(oldAgencyCode, oldSiteNumber);
+		// TODO: This might change to a new transaction type once we figure out what the new transaction file needs to look like
+		siteReport.setTransactionType("M");
+		
+		LOG.trace("Start processing primary key update transaction [" + oldAgencyCode + "-" + oldSiteNumber + " to " + newAgencyCode + "-" + newSiteNumber + "] ");
+		
+		monitoringLocation = legacyCruService.getMonitoringLocation(oldAgencyCode, oldSiteNumber, false, siteReport);
```
kmschoep-usgs, line 170: retrieves the monitoring location records to be updated by the exising ("old") primary key of agency code and site number.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
...
+		try {
+			if (!monitoringLocation.isEmpty()) {
+				
+				updatedMonitoringLocation.putAll(monitoringLocation);
```
kmschoep-usgs, line 174: basically clone the retrieved monitoring location object to create a new monitoring location object to send to the update service.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
...
+				// TODO: This might change to a new transaction type once we figure out what the new transaction file needs to look like
+				updatedMonitoringLocation.put(TRANSACTION_TYPE, "M");
+				
+				updatedMonitoringLocation.replace(AGENCY_CODE, newAgencyCode);
+				updatedMonitoringLocation.replace(SITE_NUMBER, newSiteNumber);
+
```
kmschoep-usgs, line 180: replace the agency code and site number with the new values to be updated.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
...
+			if (!monitoringLocation.isEmpty()) {
+				
+				updatedMonitoringLocation.putAll(monitoringLocation);
+				// TODO: This might change to a new transaction type once we figure out what the new transaction file needs to look like
+				updatedMonitoringLocation.put(TRANSACTION_TYPE, "M");
+				
+				updatedMonitoringLocation.replace(AGENCY_CODE, newAgencyCode);
+				updatedMonitoringLocation.replace(SITE_NUMBER, newSiteNumber);
+				
+				updatedMonitoringLocation = legacyValidatorService.doValidation(updatedMonitoringLocation, true, siteReport);
```
kmschoep-usgs, line 181: send that new monitoring location through the validator because there are some site number validations we want to make sure get run.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -150,6 +157,46 @@ public void ddotValidation(MultipartFile file) throws HystrixBadRequestException
...
+				json = mlToJson(updatedMonitoringLocation);
+				
+				json = legacyCruService.updateTransaction(updatedMonitoringLocation.get(ID), json, siteReport);
```
kmschoep-usgs, line 184: if all goes well, send the updated monitoring location object to the the updateTransaction service along with its ID so that it can be an update instead of a patch.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
kmschoep-usgs: I added some comments explaining the steps.  Does that help?

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
cschroed-usgs: This line kinda confuses me. Why would the return value of the validation method be different from the monitoring location that we pass into? Should we omit the assignment of the return value to the variable?

Alternatively, is the return value of the doValidation method best described as an updated monitoring location, or is it something different that deserves its own name? ex: `validatedMonitoringLocationUpdate` `updateValidationReport`.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
cschroed-usgs: Is this sequence possible?
1. Monitoring location is read at the top of this method
2. An attribute of the same monitoring location is modified by someone else
3. The monitoring location is sent back to the legacy cru service, updating the ID and overwriting the changes in 2.

If it is possible, do we care?

cschroed-usgs:
Thank you for the extensive tests and for updating the changelog!

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
```diff
@@ -76,6 +76,19 @@ public String updateTransaction(Object agencyCode, Object siteNumber, String jso
 		}
 	}
 	
+	public String updateTransaction(Object mlId, String json, SiteReport siteReport) {
```
zmoore-usgs, line 79: I'd lean towards having the signature here expect a string for `mlId` and put the burden of converting the object --> string on whatever calls this method.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyCruService.java
```diff
@@ -76,6 +76,19 @@ public String updateTransaction(Object agencyCode, Object siteNumber, String jso
...
+		try {
+			ResponseEntity<String> cruResp = legacyCruClient.updateMonitoringLocation(mlId.toString(), json);
+			int cruStatus = cruResp.getStatusCodeValue();
+			siteReport.addStepReport(new StepReport(SITE_UPDATE_STEP, cruStatus, 200 == cruStatus ? true : false, 200 == cruStatus ? SITE_UPDATE_SUCCESSFUL : SITE_UPDATE_FAILED));
+			return cruResp.getBody();
+		} catch (Exception e){
+			siteReport.addStepReport(new StepReport(SITE_UPDATE_STEP, HttpStatus.SC_INTERNAL_SERVER_ERROR, false, SITE_UPDATE_FAILED));
+			log.error(SITE_UPDATE_STEP + ": " + SITE_UPDATE_FAILED + ":" +  e.getMessage());
```
zmoore-usgs, line 87: I've been adding `log.error(... ,e)` to the various log messages I've come across that don't have it so that we're actually logging the exceptions that happen instead of just the short message they get. It helps add more information about the exception to the logs so we don't get stuff like `ERROR: Notification failed to send.` in the logs with no other info (which is what the notification service was logging, for example)

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
kmschoep-usgs: That is certainly possible, but if that happens we should all buy lottery tickets.  My reasoning for not being concerned: The user base for this application is small, with a limit number of users who are controlled for a relatively set group of sites.  This function would execute in a split second, so the chances that someone else would submit a ddot in the split second between the get-validation-update is very small.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
kmschoep-usgs: the doValidation service either returns the validated object that will not look different from the object passed in, or if the object doesn't validate successfully, that service throws an error.  It's convention throughout the gateway service to assign the returned the validated object back to "itself".

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
cschroed-usgs: Ok. So failure is rare. Is the cost of failure big or small? Conceptually, how does the product of the cost of failure and the frequency of failure compare to the effort of building something that is immune to this failure mode? One way: a web service `changeId(oldId, newId) -> success / fail` that runs a query like `UPDATE my_table SET id=new_id WHERE id=old_id` under the hood. Wouldn't that be pretty cheap? You may know something I don't.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
cschroed-usgs: OK. I find the convention confusing, but I think it's out of scope to address that in this pull request.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
kmschoep-usgs: I'm comfortable saying that if someone manages to update a record in the split second that someone is updating the primary key, which is already a rare occurrence, it's not really a failure, it's basically that the last thing committed to the database "wins".  We are also logging all transactions, so there would be a log of both transactions if someone noticed a discrepency.  I'm not understanding how your suggestion is a solution.  We're not updating the ID column, we're using the id column to identify the record in which we are updating the agency code and/or site number.

kmschoep-usgs: OK, I think I"m understanding what  you're saying.  Instead of sending the entire record back for update, even though we're only updating 2 columns, we could just send back for update the two columns we are actually updating.  Yes, that is certainly possible and something I can attempt.

cschroed-usgs: Yes! Well said. That's what I was trying to describe. Sorry, I wasn't clearer. I should have said "primary key" instead of "id", among other clarifications. Chalk it up to morning brain :) ?

Thanks for considering the approach! I agree that this is an edge case, so if it winds up being hard, no need to spend a ton of time on it :)

PR #115, opened 2020-04-17
MLR-496: Remove vestiges of multiple district code alerting in services and UI

kmschoep-usgs:


docker/secrets.env
```diff
@@ -1,4 +1,4 @@
-oauthClientSecret=mlr-secret
+oauthClientSecret=changeMe
```
zmoore-usgs, line 1: This new value is what the KeyCloak version of local dev expects

kmschoep-usgs: Ah, yep.  It was a bit of a merge 💩 show over here!

zmoore-usgs: Yeah sorry about that.. 😅

dpattermann-usgs:
Wait, are we getting rid of the parseWorkflow?

kmschoep-usgs:
Yes, we only needed it for the popups, and we're getting rid of the popups and replacing them with a static message in the UI.

PR #116, opened 2020-04-21
attempt to fix CORS issue in swagger ui.

kmschoep-usgs:


src/main/java/gov/usgs/wma/mlrgateway/config/WebMvcConfig.java
```diff
@@ -12,10 +14,13 @@
 
 @Configuration
 public class WebMvcConfig implements WebMvcConfigurer {
+	
+	@Value("${ui.host}")
+	private String uiDomainName;
```
cschroed-usgs, line 19: Locking this down to a specific domain rather than a wildcard should be more secure, so that is cool. If someday we want to enable other web apps to use MLR Gateway APIs, we may need to whitelist multiple domains. No change needed -- just something to keep in mind. 👍

src/main/java/gov/usgs/wma/mlrgateway/config/WebMvcConfig.java
```diff
@@ -12,10 +14,13 @@
...
 
     @Override
     public void addCorsMappings(CorsRegistry registry) {
-        registry.addMapping("/**").allowedOrigins("*").allowedMethods("GET", "POST","PUT", "DELETE","PATCH");
+        registry.addMapping("/**").allowedOrigins(uiDomainName).allowCredentials(true).allowedMethods("GET", "POST","PUT", "DELETE","PATCH");
```
cschroed-usgs, line 23: Should HEAD be added to this? OPTIONS?

kmschoep-usgs: Maybe?

kmschoep-usgs:
@zmoore-usgs , yeah, I wasn't sure after I submitted this.  I was going to try setting the uiHost to mlr-api-test.usgs.gov on the test tier just to see if that works, but I really don't know.

zmoore-usgs:
uiHost is meant to point to the Vue UI (once that's deployed), mostly for redirecting on login. We could add a configuration property for allowed CORS origins though, as Carl mentioned, and use that.

cschroed-usgs:
good point @zmoore-usgs -- if we aren't using wildcard CORS then we would need to whitelist both the `mlr-api-*` host and the ui host.

If anyone wants a refresher on CORS, MDN has a great article!
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

src/main/java/gov/usgs/wma/mlrgateway/config/WebMvcConfig.java
cschroed-usgs: My initial comment was wrong. We would need to support whitelisting multiple domains at this time to let both the UI and Swagger hit the Gateway's APIs.

PR #118, opened 2020-04-22
Revert update workflow back to submitting entire monitoring location

kmschoep-usgs:
object for update.


kmschoep-usgs:
Risk of the in-flight-record being updated while being validated is infinitesimally small, and the impact of that happening is higher, but still small, as we have an audit log tracking all changes to a record.

zmoore-usgs:
Overall looks good, just one quick question - Could we use the LegacyCRU patch workflow in order to just update one or two fields of a record without it nulling everything else out? I'm not very familiar with the differences between those two workflows on the CRU service but it seems like they should behave at least somewhat differently if they're two separate HTTP methods?

kmschoep-usgs:
No, the patch service updates the record based on matching agency code and site number, and the update service updates the record based on matching id.

PR #120, opened 2020-04-27
MLR-436: Add service that takes in a .csv file and generates transaction files

kmschoep-usgs:


pom.xml
```diff
@@ -143,6 +143,11 @@
             <artifactId>commons-io</artifactId>
             <version>2.5</version>
         </dependency>
+        <dependency>
+		  <groupId>com.opencsv</groupId>
+		  <artifactId>opencsv</artifactId>
+		  <version>5.0</version>
+		</dependency>
```
zmoore-usgs, line 150: Nitpick: Spacing looks a little off here

src/main/java/gov/usgs/wma/mlrgateway/controller/BulkTransactionFilesWorkflowController.java
```diff
@@ -0,0 +1,90 @@
...
+
+	@Operation(description="Perform the entire bulk transaction file generation workflow, sending transaction file(s) to WSC.")
+	@ApiResponses(value = {
+		@ApiResponse(responseCode = "200", description = "OK"),
+		@ApiResponse(responseCode = "400", description = "Bad Request"),
+		@ApiResponse(responseCode = "401", description = "Unauthorized"),
+		@ApiResponse(responseCode = "403", description = "Forbidden") })
+	@PreAuthorize("hasPermission(null, null)")
+	@PostMapping(path = "/bulkTransactionFiles", consumes = "multipart/form-data")
+	public UserSummaryReport bulkGenerateTransactionFilesWorkflow(@RequestPart MultipartFile file, HttpServletResponse response, Authentication authentication) {
```
zmoore-usgs, line 60: I don't have a strong opinion either way, but just wondering if it make sense to just have this as another endpoint in the existing export workflow controller vs. a new one?

zmoore-usgs: I.E: Could we re-use most of the existing export workflow controller and just have this be a "bulk" version of the endpoint that reads an uploaded CSV?

src/main/java/gov/usgs/wma/mlrgateway/util/ParseCSV.java
```diff
@@ -0,0 +1,40 @@
...
+			.withIgnoreQuotations(false)
+			.build();
+		CSVReader csvReader = new CSVReaderBuilder(reader)
+			.withSkipLines(1)
+			.withCSVParser(parser)
+			.build();
+		
+		mlList = csvReader.readAll();
+		csvReader.close();
+	    reader.close();
```
zmoore-usgs, line 35: Nitpick: Spacing is a little off here

zmoore-usgs: Also might want to put this in a try block to have it auto-close if an exception is thrown, otherwise I think reader will hang out open if an exception is thrown.

https://www.baeldung.com/java-try-with-resources

src/main/java/gov/usgs/wma/mlrgateway/workflow/BulkTransactionFilesWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+		String json;
+		//1. Parse CSV File
+		LOG.trace("Start CSV file Parsing");
+		List<String[]> monitoringLocations = new LinkedList<>();
+		try {
+			monitoringLocations = parseService.getMlList(file);
+		} catch (CsvException e1) {
+			LOG.error("An error occurred while parsing input file [" + file.getOriginalFilename() + "] ", e1);	
+		} catch (IOException e1) {
+			LOG.error("An error occurred while parsing input file [" + file.getOriginalFilename() + "] ", e1);
```
zmoore-usgs, line 59: Do we want to have any exception thrown if these errors occur? Right now I think it'd log that messages but then return back with no error has if it was just an empty file that was submitted?

src/main/java/gov/usgs/wma/mlrgateway/workflow/BulkTransactionFilesWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+	public void generateTransactionFilesWorkflow(MultipartFile file) throws HystrixBadRequestException {
+		String json;
+		//1. Parse CSV File
+		LOG.trace("Start CSV file Parsing");
+		List<String[]> monitoringLocations = new LinkedList<>();
+		try {
+			monitoringLocations = parseService.getMlList(file);
+		} catch (CsvException e1) {
+			LOG.error("An error occurred while parsing input file [" + file.getOriginalFilename() + "] ", e1);	
+		} catch (IOException e1) {
```
zmoore-usgs, line 58: Since this and the above exception do the same thing I think they can be combined into a single catch like this:

`} catch (CsvException | IOException e1) {`

src/main/java/gov/usgs/wma/mlrgateway/workflow/BulkTransactionFilesWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+			LOG.error("An error occurred while parsing input file [" + file.getOriginalFilename() + "] ", e1);
+		}
+		LOG.trace("End CSV file Parsing");
+
+		//2. Get Individual Monitoring Locations
+		LOG.trace("Start processing monitoring locations [" + file.getOriginalFilename() + "] " + monitoringLocations.size());
+			 for (String[] ml : monitoringLocations) {
```
zmoore-usgs, line 65: Nitpick: Extra indentation below this line it looks like?

src/main/java/gov/usgs/wma/mlrgateway/workflow/BulkTransactionFilesWorkflowService.java
```diff
@@ -0,0 +1,102 @@
...
+						siteReport.addStepReport(new StepReport(BULK_GENERATE_TRANSACTION_FILES_STEP, HttpStatus.SC_INTERNAL_SERVER_ERROR, false, "{\"error_message\": \"" + e.getMessage() + "\"}"));
+						BulkTransactionFilesWorkflowController.addSiteReport(siteReport);
+					}
+			 	}
+			 }
+			 LOG.trace("End processing monitoring locations [" + file.getOriginalFilename() + "] " + monitoringLocations.size());
+	}
+
+	
+	protected String mlToJson(Map<String, Object> ml) {
```
zmoore-usgs, line 89: I can't remember for sure but I think this is also in at least one other controller? If so, and if both of those controllers inherit from `BaseController` we could move this method into `BaseController` and remove it from each child class.

src/main/java/gov/usgs/wma/mlrgateway/controller/BulkTransactionFilesWorkflowController.java
kmschoep-usgs: Yeah, I went back-and-forth in head given that this service is temporary, and would it be easier removing it by deleting entire methods or going into methods and extricating that way.

zmoore-usgs: Ah that's a good point. I forgot that this is kind of a one-time use thing that will be removed post-migration. 👍

zmoore-usgs: I guess my only thought would be that perhaps there is a use-case for someone to export multiple sites at once vs. the single that our UI allows at the moment? I assume that won't be a common thing given that every new site and updated site will be pushed everywhere anyways, but might be worth asking? 🤷‍♂️ Or we just make them do it one at a time and don't worry about it haha

kmschoep-usgs: That is true.  I always assumed this would be throwaway, but perhaps there is a use case for wanting to do this occasionally.  I

kmschoep-usgs: I got confirmation from Carmen that she doesn't anticipate a need for this functionality post-post-rollout.

src/main/java/gov/usgs/wma/mlrgateway/workflow/BulkTransactionFilesWorkflowService.java
```diff
@@ -0,0 +1,115 @@
...
+	public void generateTransactionFilesWorkflow(MultipartFile file) throws HystrixBadRequestException {
+		String json;
+		//1. Parse CSV File
+		LOG.trace("Start CSV file Parsing");
+		List<String[]> monitoringLocations = new LinkedList<>();
+		try {
+			monitoringLocations = parseService.getMlList(file);
+		} catch (CsvException | IOException e) {
+			LOG.error("An error occurred while parsing input file [" + file.getOriginalFilename() + "] ", e);	
+			if (e instanceof FeignBadResponseWrapper) {
```
zmoore-usgs, line 66: In this block I don't think we can get a FeignBadResponseWrapper because we're only catching CsvException and IOException. Could add FeignBadResponseWrapper to that catch list though

kmschoep-usgs: OK.  My exception handling is weak.  Thank you for your patience.  😁

PR #122, opened 2020-04-30
MLR-453: Update gateway update primary key workflow to call new file

kmschoep-usgs:
export change api.


src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
```diff
@@ -140,6 +140,7 @@ public UserSummaryReport updatePrimaryKeyWorkflow(
 			@RequestParam String newAgencyCode,
 			@RequestParam String oldSiteNumber,
 			@RequestParam String newSiteNumber,
+			@RequestParam String reasonText,
```
zmoore-usgs, line 143: Do we need to do any sanitizing of the reason text input or put any limitations (like a character limit) on it?

kmschoep-usgs: I thought of that and figured it could be handled in the UI, but I can implement something in the service that at least enforces alpha-numeric inputs.

zmoore-usgs: Sounds good to me. There are a lot of JavaX Validation annotations that make those sorts of validations pretty easy:

1. @Pattern - https://docs.oracle.com/javaee/7/api/javax/validation/constraints/Pattern.html (https://stackoverflow.com/questions/17481029/pattern-for-alphanumeric-string-bean-validation)
2. @Size - https://docs.oracle.com/javaee/6/api/javax/validation/constraints/Size.html

zmoore-usgs: Woah, not sure what the `@` annotations are doing to the markup to make the capitalization all wonky but I definitely typed those as `@Pattern` and `@Size` (and just tried to edit the comment to make sure). Weird 😄

zmoore-usgs: Oh it's tagging GitHub users apparently. Whoops!

src/main/java/gov/usgs/wma/mlrgateway/util/ValidateTextInput.java
```diff
@@ -0,0 +1,28 @@
...
+public class ValidateTextInput {
+	private Logger log = LoggerFactory.getLogger(ValidateTextInput.class);
+	public static final String INTERNAL_ERROR_MESSAGE = "Invalid characters submitted in reasonText. Only alpha-numeric characters are allowed.";
+	
+	public String validateInput(String StrInput) {
+		Boolean valid = true;
+		if (!StrInput.isEmpty()) {
+			valid = StrInput.matches("^[a-zA-Z0-9 ]*$");
+		}
+		if (!valid) {
```
zmoore-usgs, line 20: This would allow empty strings through I think. Is it okay to have no reason provided? That might be more of a question for Joe/Carmen

src/main/java/gov/usgs/wma/mlrgateway/util/ValidateTextInput.java
```diff
@@ -0,0 +1,28 @@
...
+
+import gov.usgs.wma.mlrgateway.FeignBadResponseWrapper;
+
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ValidateTextInput {
```
zmoore-usgs, line 11: I think this sort of validation could be done via annotations (see my comment from earlier) but I don't have a huge preference between that and doing it this way.

src/main/java/gov/usgs/wma/mlrgateway/util/ValidateTextInput.java
```diff
@@ -0,0 +1,28 @@
...
+	public static final String INTERNAL_ERROR_MESSAGE = "Invalid characters submitted in reasonText. Only alpha-numeric characters are allowed.";
+	
+	public String validateInput(String StrInput) {
+		Boolean valid = true;
+		if (!StrInput.isEmpty()) {
+			valid = StrInput.matches("^[a-zA-Z0-9 ]*$");
+		}
+		if (!valid) {
+			log.error(INTERNAL_ERROR_MESSAGE);
+			throw new FeignBadResponseWrapper(HttpStatus.SC_INTERNAL_SERVER_ERROR, null, INTERNAL_ERROR_MESSAGE);
```
zmoore-usgs, line 22: Perhaps have this return an HTTP 400 instead of a 500 since it'd be caused by a user making an invalid request rather than a service error?

src/main/java/gov/usgs/wma/mlrgateway/util/ValidateTextInput.java
kmschoep-usgs: Yeah, was in the middle of this before I saw your annotations comment, I'm trying that now.

PR #124, opened 2020-05-04
send returned updated monitoring location object to the file exporter.

kmschoep-usgs:


src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -230,4 +231,19 @@ protected String mlToJson(Map<String, Object> ml) {
...
+	protected Map<String, Object> jsonToMl(String json) {
+		ObjectMapper mapper = new ObjectMapper();
+		Map<String, Object> ml = new HashMap<>();
+		
+		try {
+			ml = mapper.readValue(json, Map.class);
+			
+		} catch (Exception e) {
+			// Unable to determine when this might actually happen, but the api says it can...
+			throw new FeignBadResponseWrapper(HttpStatus.SC_INTERNAL_SERVER_ERROR, null, "{\"error_message\": \"Unable to serialize transformer output.\"}");
```
zmoore-usgs, line 245: Might want to change the error message here a bit to something like "Unable to deserialize legacy CRU output." and also consider logging the raw response JSON at the error level so we can see what the response was when looking at the logs.

PR #125, opened 2020-05-06
Enable Refresh Token

zmoore-usgs:
This PR ended up including a lot of changes which I can go into the reasoning behind in further detail if anybody wants. Probably will go over this at the next team demo. I learned a whole lot through this! Also there's definitely a few white-space things and typos to cleanup which I'll get to tomorrow.

**At a high level:**

_1. The Gateway itself no longer accepts JWT tokens as authentication._

It has gone back to using HTTP Session for Authentication, however rather than using a session cookie it uses a session HTTP Header (X-Auth-Token) which allows us to use the same pattern as before for login (redirect to Gateway --> redirect to KeyCloak --> return to Gateway and create authenticated HTTP session --> return session token to UI via a query parameter in the redirect back to the UI --> use session token in HTTP header for future requests.

The main reason for this is that JWTs are, by design, stateless which means that they can't be "refreshed" via the refresh token OAuth2 flow like normal OAuth2 tokens can. When you make a request to a service with a JWT in the Authorization header that sets up a stateless "session" in that service based solely on the token itself, and we don't want tokens to be able to just refresh themselves so the JWT can't be refreshed. Instead, we store the JWT that KeyCloak issues in the HTTP session along with the refresh token. Since that is managed server-side it can be refreshed automatically before it expires using the long-lived refresh token that we don't want to hand out to clients. So, by using HTTP session for requests to the Gateway we get both the benefits of using JWT for downstream services (when you run a workflow via the Gateway your JWT is extracted from your session and sent to the downstream services) and HTTP session for clients to allow us to refresh the token. Additionally, by using X-Auth-Token instead of a traditional session cookie we can pass the session around across different domains and store it in more secure areas like session storage.

This article was very helpful for understanding all of this: https://dzone.com/articles/stop-using-jwts-as-session-tokens

There is a related PR to the MLR user interface to address this change: https://github.com/USGS-CIDA/MLR-User-Interface/pull/22

_2. The username is pulled from a different token claim_ 

The username was previously extracted from the "sub" field of the OAuth2/JWT token. This worked fine in WaterAuth as the username was equivalent to the user's unique ID in WaterAuth. In KeyCloak however users each have a UUID which is sent as the "sub" claim of the token and this is not equivalent to their username. I worked around this previously using a KeyCloak mapper to override the "sub" claim with the user's username, which mimicked the behavior of WaterAuth. This worked fine until Refresh Tokens. Refresh tokens link to their Access Token and to the authorized user in KeyCloak via the "sub" claim, and for Refresh Tokens and KeyCloak itself that's the user's UUID. By overriding the "sub" claim in the Access Token with the username that link was broken and refresh tokens didn't work properly. So, we'll need to disable that mapper and do a more "normal" thing in KeyCloak and map the username to a new, custom username claim. I used KeyCloak's default claim "preferred_username" for that and configured the Gateway to pull the username from that claim instead of "sub".

There is a related PR for MLR Legacy CRU (the only other MLR service that pulls the username from the token) to address this change: https://github.com/USGS-CIDA/MLR-Legacy-Service/pull/91

_3. MLR Local Dev_

There will also be a follow-up PR for MLR Local Dev, but I'll get around to that tomorrow as I should really stop working now and go make dinner and walk Olive... 😅


src/main/resources/application.yml
```diff
@@ -15,6 +15,10 @@ security:
   basic:
     enabled: false
   maintenanceRoles: ${maintenanceRoles:}
+  token:
+    claims:
+      email: email
+      username: preferred_username
```
dpattermann-usgs, line 21: So is this used to get the usernames from whatever comes back from sciencebase?

zmoore-usgs: Yep, these two config properties are used in UserAuthUtils to pull values from the token. The values here are the names of the claims (JSON properties) of the token to pull from.

cschroed-usgs:
I appreciate the time you put into implementing this complex change and into writing the prose that describes the context for this pull request. Personally, I have a hard time understanding this new proposed auth workflow without diagrams. Are there any existing diagrams out there on the webternets that match or closely resemble the approach we're switching to?

src/main/java/gov/usgs/wma/mlrgateway/config/OAuth2Config.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+	
+	@Bean
+	public OAuth2AuthorizedClientManager authorizedClientManager(
+		ClientRegistrationRepository clientRegistrationRepository,
+		OAuth2AuthorizedClientRepository authorizedClientRepository
+	) {
+		OAuth2AuthorizedClientProvider authorizedClientProvider = 
+			OAuth2AuthorizedClientProviderBuilder.builder()
+				.refreshToken(configurer -> configurer.clockSkew(accessTokenExpireSkew))
```
cschroed-usgs, line 54: I haven't played with [method references](https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html) in a while, but could they be used to simplify these lines?


Something like
`.refreshToken(Configurer:clockSkew(accessTokenExpireSkew)`

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2Filter.java
```diff
@@ -0,0 +1,56 @@
...
+		return !(ctx.containsKey(FORWARD_TO_KEY) && !ctx.get(FORWARD_TO_KEY).toString().isEmpty())
+				&& ctx.getOrDefault(SERVICE_ID_KEY, "").toString().startsWith(MLR_SERVICE_PREFIX);
+	}
+
+	@Override
+	public Object run() {
+		RequestContext ctx = RequestContext.getCurrentContext();
+		String authToken = userAuthUtil.getTokenValue(SecurityContextHolder.getContext().getAuthentication());
+
+		if(authToken != null && !authToken.isEmpty() && !ctx.getZuulRequestHeaders().containsKey("Authorization")) {
```
cschroed-usgs, line 37: If these conditions are not satisfied, should anything be logged or thrown?

zmoore-usgs:
This is the best one I've found and the one I referred to a lot, with the relevant replacements being:

Browser = MLR UI in client browser
Grails     = MLR Gateway
Google  = WMA KeyCloak


What this doesn't include is the refresh token flow, which is this:

https://www.tutorialspoint.com/oauth2.0/images/refreshing_an_expired_access_token.jpg

Where the relevant replacements are:

Client                         = MLR Gateway
Resource Server         = MLR Services behind the Gateway
Authorization Server = WMA KeyCloak

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2Filter.java
zmoore-usgs: It should be kind of inherently - The downstream service will reject the request with a 401 and complain there was no access token provided with the request, which should be captured in the response logs for the client.

cschroed-usgs: I'm not sure what the Zuul Filter conventions are, but based solely on the method names, I'm wondering if it would make more sense if the auth token lookup and this condition were moved to the `shouldFilter` method?

src/main/java/gov/usgs/wma/mlrgateway/config/OAuth2Config.java
zmoore-usgs: Hmm I can take a look at that, I just copied most of this bit from the Spring docs

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2Filter.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
+
+import gov.usgs.wma.mlrgateway.util.UserAuthUtil;
+
+import com.netflix.zuul.ZuulFilter;
+import com.netflix.zuul.context.RequestContext;
+
+@Component
+public class ZuulOAuth2Filter extends ZuulFilter {
```
cschroed-usgs, line 18: Should we include an automated test to ensure that requests continue to have the correct Authorization header attached to them in the future?

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2Filter.java
zmoore-usgs: I'm not entirely sure, but I based this on a similar one that Dave S. wrote for AQCU: https://github.com/USGS-CIDA/aqcu-gateway/blob/master/src/main/java/gov/usgs/aqcu/CustomZuulFilter.java

My understanding is "shouldFilter" is meant to decide if the "run" method should be run at all. I think we could move the check for whether the request headers already have a defined "Authorization" header to that function for sure. As far as the authToken check I'm not as sure. I don't think we can pass the token from `shouldFilter()` to `run()` so we'd have to look it up in both places if we move that check there.

cschroed-usgs: 👍  Good point on the token lookup, might as well keep that here. I like the idea of moving the check for existing Authorization headers over to `shouldFilter`, unless some sort of Zuul convention says that kind of check isn't legit to do there.

src/main/java/gov/usgs/wma/mlrgateway/controller/UtilController.java
```diff
@@ -7,15 +7,20 @@
...
-import org.apache.http.HttpStatus;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.context.request.RequestContextHolder;
```
cschroed-usgs, line 23: The two methods in this class both relate to auth. Should this class be renamed to something like AuthController? Should the request mapping change to something like "/auth"?

zmoore-usgs: Seems reasonable to me. I think at one point we had ambitions of using this controller for other utility stuff but that never came to fruition so renaming it makes sense now

src/main/java/gov/usgs/wma/mlrgateway/config/OAuth2Config.java
zmoore-usgs: I'm not sure that works here since we're not just passing in a method to run, but rather running a method against an instance reference and providing it with an argument. Seems like method references are more for passing an implementation to a function call or something? Where in this case where we're executing a setter on an instance reference with a specific value.

But I might be missing something?

src/test/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2FilterTest.java
```diff
@@ -0,0 +1,111 @@
...
+    public void runFilterNoTokenTest() {
+        when(userAuthUtil.getTokenValue(any())).thenReturn(null);
+        RequestContext context = new RequestContext();
+        context.set(SERVICE_ID_KEY, "mlrLegacyCru");
+        RequestContext.testSetCurrentContext(context);
+        assertNull(RequestContext.getCurrentContext().getZuulRequestHeaders().get(ZuulOAuth2Filter.AUTHORIZATION_HEADER));
+        filter.run();
+        assertNull(RequestContext.getCurrentContext().getZuulRequestHeaders().get(ZuulOAuth2Filter.AUTHORIZATION_HEADER));
+    }
+}
```
cschroed-usgs, line 111: Great tests 👍 !

src/main/java/gov/usgs/wma/mlrgateway/config/OAuth2Config.java
cschroed-usgs: Allrighty! Thanks for exploring that!

PR #127, opened 2020-05-07
Tweak default cookie serializer SameSite value to LAX.

zmoore-usgs:


zmoore-usgs:
Going to go ahead and merge after this passes Travis since it's a trivial change, explanation provided in Teams.

PR #128, opened 2020-05-09
Enable configurable session timeout for Spring Session…

zmoore-usgs:
…MapSessionRepository.

Turns out when using Spring Session and the MapSessionRepository the default spring way of overriding session timeout which we had setup before (`spring.session.timeout` config parameter) is not used, it has to be manually set on the MapSessionRepository instance. When not set the default is 30 minutes which is longer than I had my session sit when testing locally and was resulting in the login timeouts Joe was still seeing on Test.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
@@ -2,6 +2,7 @@ package gov.usgs.wma.mlrgateway.config;
 
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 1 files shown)


zmoore-usgs:
Given the small number of changes here I'm going to merge once Travis passes so that it can be tested on Test soon (to make sure this was the culprit and not some sort of session timeout being enforced by AWS or something), but feel free to still review and I can make changes if anything looks like it needs changing.

src/main/java/gov/usgs/wma/mlrgateway/config/WebSecurityConfig.java
```diff
@@ -15,6 +16,9 @@
 @EnableSpringHttpSession
 @Configuration
 public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
+
+	@Value("${server.session.timeout:}")
```
zmoore-usgs, line 20: This was already configured in the application.yml since it's the property we use to configure the session timeout when _not_ using SpringSession like we are now.

PR #129, opened 2020-05-12
Mlr 495

kmschoep-usgs:


kmschoep-usgs:
Any ideas on how to make the guts of the new service testable are welcome.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -95,6 +95,59 @@ public LegacyValidatorService(LegacyCruService legacyCruService, LegacyValidator
...
+				siteReport.addStepReport(new StepReport(SITE_VALIDATE_STEP, duplicateValidationStatus, false, clientErrorParser.parseClientError("LegacyCruClient#validateMonitoringLocation(String)", ((FeignBadResponseWrapper)e).getBody())));
+			} else {
+				log.error("An error occured during LegacyCRU Duplicate Validation: ", e);
+				duplicateValidationStatus = HttpStatus.SC_INTERNAL_SERVER_ERROR;
+				siteReport.addStepReport(new StepReport(SITE_VALIDATE_STEP, duplicateValidationStatus, false, e.getMessage()));
+			}
+		}
+		try {
+			ResponseEntity<String> validationResponse;
+			//  Need to implement this method as if it were an add transaction, since we want it to fail if a monitoring location exists with the new primary key
```
zmoore-usgs, line 118: Does "this method" refer to the `attachExistingMonitoringLocation` call below this line or this entire `try{}` block? A little bit confusing to follow the comments here since this talks about implementing the method as an add transaction but then the next comment talks about needing to validate as if it's an update transaction.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -95,6 +95,59 @@ public LegacyValidatorService(LegacyCruService legacyCruService, LegacyValidator
...
+			validationResponse = legacyValidatorClient.validateUpdate(validationPayload);
+			ml = verifyValidationStatus(ml, validationResponse);
+			siteReport.addStepReport(new StepReport(VALIDATION_STEP, validationResponse.getStatusCodeValue(), true, "{\"validator_message\": " + validationResponse.getBody() + "}" ));
+		} catch (Exception e) {
+			if(e instanceof FeignBadResponseWrapper) {
+				log.debug("An error occured during LegacyValidator Validation: ", e);
+				otherValidationStatus = ((FeignBadResponseWrapper)e).getStatus();
+				siteReport.addStepReport(new StepReport(VALIDATION_STEP, otherValidationStatus, false, ((FeignBadResponseWrapper)e).getBody()));
+			} else {
+				log.error("An error occured during LegacyValidator Validation: ", e);
```
zmoore-usgs, line 131: Might note in these logs somewhere that this was during a PK update so we can distinguish validation errors between those and DDot file workflows. Maybe something like `An error occurred during PK update LegacyValidator Validation: `?

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
```diff
@@ -95,6 +95,59 @@ public LegacyValidatorService(LegacyCruService legacyCruService, LegacyValidator
...
+		int otherValidationStatus = 200;
+		try {
+			doDuplicateValidation(ml, siteReport);
+		} catch (Exception e) {
+			if(e instanceof FeignBadResponseWrapper) {
+				log.debug("An error occured during LegacyCRU Duplicate Validation: ", e);
+				duplicateValidationStatus = ((FeignBadResponseWrapper)e).getStatus();
+				siteReport.addStepReport(new StepReport(SITE_VALIDATE_STEP, duplicateValidationStatus, false, clientErrorParser.parseClientError("LegacyCruClient#validateMonitoringLocation(String)", ((FeignBadResponseWrapper)e).getBody())));
+			} else {
+				log.error("An error occured during LegacyCRU Duplicate Validation: ", e);
```
zmoore-usgs, line 111: (Same as earlier comment re: logs)

zmoore-usgs:
Will take a look at testing stuff

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
kmschoep-usgs: Sorry, yes, that comment only applies to `attachExistingMonitoringLocation`.

src/main/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorService.java
kmschoep-usgs: Ah, yep. good call.

zmoore-usgs:
The current test file for the validation service is here: https://github.com/USGS-CIDA/MLR-Gateway/blob/master/src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java

If you follow those patterns and test something like:

```
1. Success case where both validation clients are mocked to return 200
2. Error case where legacyCru duplicate validator returns expected validation error
3. Error case where legacyCru duplicate validator returns unexpected exception
4. Error case where legacyValidator returns expected validation error
5. Error case where legacyValidator returns unexpected exception
```
Then in each of those tests ensure that the right thing is returned (or right exception thrown) and that the Gateway report includes the proper expected section.

I.E:

```
1. Test that a monitoring location is returned (like this one: https://github.com/USGS-CIDA/MLR-Gateway/blob/master/src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java#L91)
2. Wrap the call in a try-catch and include a fail clause in the try block after the invocation (like this one: https://github.com/USGS-CIDA/MLR-Gateway/blob/master/src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java#L194)
```

That should cover it I think?

kmschoep-usgs:
Ah, yep!  I'm too close to the code, totally missed that testing package.

src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java
```diff
@@ -420,4 +420,271 @@ public void validatorService_doValidation_DuplicateValidationFailure_InvalidData
...
+			fail("Validation should throw an exception when errors are found in order to prevent further processing of this transaction.");
+		} catch (FeignBadResponseWrapper e) {
+			assertTrue(e.getStatus() == 400);
+			assertTrue(LEGACY_VALIDATION_ERROR_BODY.equals(e.getBody()));
+		}
+		
+		GatewayReport gatewayReport = WorkflowController.getReport();
+		gatewayReport.addSiteReport(siteReport);
+		
+		assertEquals(siteReport.getSteps().size(), 2);
```
cschroed-usgs, line 470: Thanks for implementing extensive tests! Some of the parameters to the assertion methods in this pull request are in the order `actual`, `expected`. [JUnit](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertEquals-java.lang.Object-java.lang.Object-) expects the parameters to be in the order `expected`, `actual`, and it generates messages accordingly when tests fail. Could the order of the parameters to the assertions in this pull request please be made consistent with the JUnit API so that developers can quickly understand and address test failures in the future?

It looks like some nearby methods that were not touched in this pull request also put parameters to assertion methods in an unconventional order, but I don't think it's super important to resolve that now.

src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java
```diff
@@ -420,4 +420,271 @@ public void validatorService_doValidation_DuplicateValidationFailure_InvalidData
...
+		given(legacyCruService.getMonitoringLocation(anyString(), anyString(), anyBoolean(), any())).willReturn(ml);
+		given(legacyCruService.validateMonitoringLocation(any(), any())).willReturn("{}");
+		given(legacyValidatorClient.validateUpdate(anyString())).willReturn(validatorResponse);
+		
+		try{
+			service.doPKValidation(ml, siteReport);
+			fail("Validation should throw an exception when errors are found in order to prevent further processing of this transaction.");
+		} catch (FeignBadResponseWrapper e) {
+			assertTrue(e.getStatus() == 500);
+			assertTrue(LEGACY_VALIDATION_ERROR_BODY.equals(e.getBody()));
```
cschroed-usgs, line 524: Could these `assertTrue`s be switched to `assertEquals`?  `assertEquals` tends to provide more helpful feedback (i.e. a string diff) than `assertTrue` and `assertFalse`. Are there other assertions in this pull request that would benefit from this conversion?

src/test/java/gov/usgs/wma/mlrgateway/service/LegacyValidatorServiceTest.java
kmschoep-usgs: Yeah, I've noticed that too, but have been inconsistent with fixing them.

PR #131, opened 2020-05-14
Another attempt to get Bisonconnect Chrome to upload a file

kmschoep-usgs:
https://stackoverflow.com/questions/46563182/spring-file-upload-required-request-part-is-not-present

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
@@ -4,6 +4,7 @@ import java.time.Clock;
 import java.util.Collections;
 import java.util.Comparator;
 
+import javax.servlet.MultipartConfigElement;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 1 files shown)


src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
```diff
@@ -48,6 +50,11 @@
 	public static final String PRIMARY_KEY_UPDATE_WORKFLOW_SUBJECT = "Submitted Primary Key Update Transaction";
 	private final Clock clock;
 	
+	@Bean
+    public MultipartConfigElement multipartConfigElement() {
```
zmoore-usgs, line 54: Nitpick: Tabs vs. spaces

kmschoep-usgs: whoops.  We'll see if this even works first.  I can't reproduce this locally to test, so I'm just throwing 🍝  right now.

PR #132, opened 2020-05-14
Update to Spring Boot 2.3.0 for RefreshToken expiration handling

zmoore-usgs:
- Refactored UserAuthUtils and OAuth2Config to account for new pre-configured beans in Spring Boot 2.3.0
- Added custom Zuul Error Filter to handle expired token errors when routing requests through Zuul
- Cleaned up expired token error handling in the GlobalDefaultExceptionHandler so that it properly invalidates the session when the refresh token is expired (fixes issue Dominick had)
- Added explicit checks to validate the token to controller methods because with the way we have Feign configured OAuth2 exception generated from the Feign OAuth2 filter weren't being passed all the way back to the GlobalDefaultExceptionHandler
- Limited the scopes requested by the OAuth2 Authorization Request since Spring Boot 2.3.0 by default requests all available scopes (and results in big tokens from KeyCloak which has lots of optional scope data)
- Added default logout redirect URL wrapper to immediately direct the user back to login since MLR doesn't support any un-authenticated users. The wrapper is needed because by default Spring Security makes whatever URL you put as the logout success URL not require any auth (which makes sense) but that means it can't just send you right to the login URL which does require auth (in order to trigger the login)
- Added additional tests


pom.xml
```diff
@@ -38,14 +38,19 @@
             <name>Maven Snapshots and Releases</name>
             <url>https://artifactory.wma.usgs.gov/artifactory/maven-central</url>
         </repository>
+        <repository> 
+            <id>repository.spring.milestone</id> 
+            <name>Spring Milestone Repository</name> 
+            <url>https://repo.spring.io/milestone</url> 
+        </repository>
```
zmoore-usgs, line 45: The proper 2.3.0 release was cut this morning but has since been retracted due to a regression with the build (not related to the content of Spring Boot itself) https://github.com/spring-projects/spring-boot/issues/21451 

Hopefully it will be out again soon, but for now this is on the Release Candidate

pom.xml
```diff
@@ -99,11 +104,11 @@
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
```
zmoore-usgs, line 102: No longer need this dependency since the Gateway is now just an OAuth2 Client and not a Resource Server

pom.xml
```diff
@@ -99,11 +104,11 @@
...
+            <artifactId>spring-boot-starter-actuator</artifactId>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-actuator</artifactId>
+            <artifactId>spring-boot-starter-validation</artifactId>
```
zmoore-usgs, line 111: With Spring Boot 2.3.0 validation is not included in spring-boot-starter-web and must be explicitly included

src/main/java/gov/usgs/wma/mlrgateway/config/OAuth2Config.java
```diff
@@ -49,20 +43,10 @@ public OAuth2AuthorizedClientManager authorizedClientManager(
 		ClientRegistrationRepository clientRegistrationRepository,
 		OAuth2AuthorizedClientRepository authorizedClientRepository
 	) {
-		OAuth2AuthorizedClientProvider authorizedClientProvider =
```
zmoore-usgs, line 52: In Spring Boot 2.3.0 there is a Bean defined for this by default

src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
```diff
@@ -111,6 +112,7 @@ public UserSummaryReport legacyWorkflow(@RequestPart MultipartFile file, HttpSer
 		@ApiResponse(responseCode = "403", description = "Forbidden") })
 	@PostMapping(path = "/ddots/validate", consumes = "multipart/form-data")
```
kmschoep-usgs, line 113: @zmoore-usgs , I recently added the `consumes = 'mulitpart/form-data'` to enable the new swagger ui be able to have the browse file widget available.  If this merge doesn't magically resolve this issue, removing this might be the next thing to try.

src/test/java/gov/usgs/wma/mlrgateway/controller/GlobalDefaultExceptionHandlerTest.java
```diff
@@ -34,54 +47,91 @@ public void setUp() throws Exception {
...
-	}	
+	}
+
+	@Test
+	public void handleOAuth2AuthorizationException() throws IOException, ServletException {
+		HttpServletResponse response = new MockHttpServletResponse();
+		HttpServletRequest servRequest = mock(HttpServletRequest.class);
+		HttpSession session = mock(HttpSession.class);
+		when(servRequest.getSession()).thenReturn(session);
+		String expected = "Your login has expired.";
```
cschroed-usgs, line 113: Sometimes it is helpful to source these values from a `public static final String` on the implementation class. I only spy two usages of the message in tests in the pull request though, so not sure it's worth it here 🤷 Up to you!

src/test/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2ErrorFilterTest.java
```diff
@@ -0,0 +1,148 @@
...
+    UserAuthUtil userAuthUtil;
+
+    @Test
+    public void shouldFilterHappyPathTest() {
+        RequestContext context = new RequestContext();
+        
+        ClientAuthorizationRequiredException clientAuthException = new ClientAuthorizationRequiredException("test");
+        when(exceptionWrapper.getCause()).thenReturn(clientAuthException);
+        context.setThrowable(exceptionWrapper);
+        RequestContext.testSetCurrentContext(context);
```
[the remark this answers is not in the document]
zmoore-usgs: Yeah, Zuul is actually currently in maintenance mode now so that shouldn't be going away at this point and it is their recommended way of doing it but I agree it's really weird.

Whenever we get around to the next MLR project we should probably look into dropping Zuul altogether since it's not really developed anymore (beyond bug fixes and security updates). I'm not sure what the intended replacement is.

src/main/java/gov/usgs/wma/mlrgateway/util/UserAuthUtil.java
```diff
@@ -3,12 +3,12 @@
...
+import org.springframework.security.oauth2.client.ClientAuthorizationRequiredException;
 import org.springframework.security.oauth2.client.OAuth2AuthorizeRequest;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClient;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientManager;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
 import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
-import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.stereotype.Component;
 
 @Component
```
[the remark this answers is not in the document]
zmoore-usgs: Probably so, yeah

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2ErrorFilter.java
```diff
@@ -0,0 +1,48 @@
...
+ * Converts a 500 zuul exception to a 401 in the case where the error was caused
+ * by an OAuth2AuthorizationException.
+ *
+ */
+@Component
+public class ZuulOAuth2ErrorFilter extends SendErrorFilter {
+
+	@Override
+	public boolean shouldFilter() {
+		return super.shouldFilter() && findZuulException(RequestContext.getCurrentContext().getThrowable()).getThrowable().getCause() instanceof OAuth2AuthorizationException;
```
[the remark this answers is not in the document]
zmoore-usgs: I'm not sure either, but will dig into it and see. I think `super.shouldFilter()` might ensure that `findZuulException(...)` is never null, but will check.

src/main/java/gov/usgs/wma/mlrgateway/controller/GlobalDefaultExceptionHandler.java
```diff
@@ -38,18 +40,18 @@
 	static final String ERROR_MESSAGE_KEY = "error_message";
 
 	@ExceptionHandler(Exception.class)
-	public @ResponseBody Map<String, String> handleUncaughtException(Exception ex, WebRequest request, HttpServletResponse response) throws IOException {
+	public @ResponseBody Map<String, String> handleUncaughtException(Exception ex, WebRequest request, HttpServletRequest servletRequest, HttpServletResponse servletResponse) throws IOException {
```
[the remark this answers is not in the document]
zmoore-usgs: Yeah, I think it's worth switching to those at some point. Right now given how much the two lower level classes are used throughout the Gateway code I'm not sure I want to bundle that in this PR but I can make a follow-up improvement task in JIRA to look into that.

src/main/java/gov/usgs/wma/mlrgateway/config/ZuulOAuth2ErrorFilter.java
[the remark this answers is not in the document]
zmoore-usgs: I verified that if super.shouldFilter() is true in this case findZuulException cannot be null

src/test/java/gov/usgs/wma/mlrgateway/util/UserAuthUtilTest.java
```diff
@@ -0,0 +1,274 @@
...
+        );
+
+        try {
+            userAuthUtil.getRefreshAuthorizedClient(mockOAuthToken);
+            fail("Expected ClientAuthorizationRequiredException but got no exception.");
+        } catch(ClientAuthorizationRequiredException e) {
+            assertTrue(e.getMessage().contains(TEST_CLIENT_ID));
+        } catch(Exception e) {
+            fail("Expected ClientAuthorizationRequiredException but got " + e.getClass().getName());
+        }
```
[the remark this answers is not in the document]
zmoore-usgs: I'm going to leave the test pattern alone for now because the current pattern is used pretty extensively throughout the tests. Will create a follow-up improvement task to look into annotation based exception testing.

src/main/java/gov/usgs/wma/mlrgateway/util/UserAuthUtil.java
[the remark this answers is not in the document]
zmoore-usgs: I made this change

src/test/java/gov/usgs/wma/mlrgateway/controller/GlobalDefaultExceptionHandlerTest.java
zmoore-usgs: Made this change

pom.xml
zmoore-usgs: It's out now

zmoore-usgs:
Going to go for the merge now to see if this has any impact on the weird issue on Test

PR #134, opened 2020-05-15
update how the update primary key workflow adds sites and steps to the

kmschoep-usgs:
gateway report.


src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -167,13 +167,25 @@ public void updatePrimaryKeyWorkflow(String oldAgencyCode, String oldSiteNumber,
...
+		SiteReport newSiteReport = new SiteReport(newAgencyCode, newSiteNumber);
 		// TODO: This might change to a new transaction type once we figure out what the new transaction file needs to look like
-		siteReport.setTransactionType("M");
+		oldSiteReport.setTransactionType("M");
+		newSiteReport.setTransactionType("M");
 		
 		LOG.trace("Start processing primary key update transaction [" + oldAgencyCode + "-" + oldSiteNumber + " to " + newAgencyCode + "-" + newSiteNumber + "] ");
 		
-		monitoringLocation = legacyCruService.getMonitoringLocation(oldAgencyCode, oldSiteNumber, false, siteReport);
+		try {
```
kmschoep-usgs, line 178: This basically treats the fetch of the old and the validation of the new as 2 transactions.  There are pros and cons.  The pro is that the output report is more clear about what agency code-site number combo is throwing an error.  The con is that the gateway report shows this as 2 transactions, which might be confusing.

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -208,18 +220,21 @@ public void updatePrimaryKeyWorkflow(String oldAgencyCode, String oldSiteNumber,
...
+				WorkflowController.addSiteReport(newSiteReport);
 			} else {
 				LOG.error("An error occurred while processing primary key update transaction [" + oldAgencyCode + "-" + oldSiteNumber + " to " + newAgencyCode + "-" + newSiteNumber + "] ", e);
-				siteReport.addStepReport(new StepReport(PRIMARY_KEY_UPDATE_TRANSACTION_STEP, HttpStatus.SC_INTERNAL_SERVER_ERROR, false, "{\"error_message\": \"" + e.getMessage() + "\"}"));
+				newSiteReport.addStepReport(new StepReport(PRIMARY_KEY_UPDATE_TRANSACTION_STEP, HttpStatus.SC_INTERNAL_SERVER_ERROR, false, "{\"error_message\": \"" + e.getMessage() + "\"}"));
+				WorkflowController.addSiteReport(newSiteReport);
 			}
 		}
-		WorkflowController.addSiteReport(siteReport);
+		WorkflowController.addSiteReport(oldSiteReport);
```
dpattermann-usgs, line 237: I think it makes sense to split them up into two pieces. One question though: Should this addSiteReport be moved to the oldSiteReport get block? Similar to how the newSiteReport is added in it's try catch block

kmschoep-usgs: 🤔  I can try that.  Maybe that would eliminate it reporting 2 transactions instead of one.  I'll have to see how it reports out successes and failures.

dpattermann-usgs: Yeah, just a thought. I just wasn't sure about why you put the old one here but the new one in the try.

kmschoep-usgs: It was so if the old site failed to be fetched in the first try, the workflow report wouldn't show the new site as a success, which would be false, as it wouldn't even get that far.

PR #137, opened 2020-05-20
feature-toggle the primary key update functionality.

kmschoep-usgs:


src/main/resources/application.yml
```diff
@@ -49,6 +49,8 @@ notification:
   email:
     cc-list: ${emailCCList:}
 
+enablePrimaryKeyUpdate: ${enablePrimaryKeyUpdate:false}
```
zmoore-usgs, line 52: Since the variable name and substitution name are the same here I think this can just be `enablePrimeyKeyUpdate: false` and then it can still be overridden with the environment variable `enablePrimaryKeyUpdat`

src/main/java/gov/usgs/wma/mlrgateway/controller/WorkflowController.java
```diff
@@ -155,38 +162,41 @@ public UserSummaryReport updatePrimaryKeyWorkflow(
...
-			if (e instanceof FeignBadResponseWrapper) {
-				int status = ((FeignBadResponseWrapper) e).getStatus();
-				WorkflowController.addWorkflowStepReport(new StepReport(LegacyWorkflowService.PRIMARY_KEY_UPDATE_WORKFLOW_FAILED, status, false, ((FeignBadResponseWrapper) e).getBody()));
-			} else {
-				int status = HttpStatus.SC_INTERNAL_SERVER_ERROR;
-				WorkflowController.addWorkflowStepReport(new StepReport(LegacyWorkflowService.PRIMARY_KEY_UPDATE_WORKFLOW_FAILED, status, false, e.getLocalizedMessage()));
+		if (!enablePrimaryKeyUpdate) {
+			throw new UnsupportedOperationException("Feature not enabled");
+		   }
+			userAuthService.validateToken(authentication);
```
zmoore-usgs, line 168: Nitpick: Indentation looks a little funky here

src/test/java/gov/usgs/wma/mlrgateway/controller/WorkflowControllerTest.java
```diff
@@ -227,6 +228,21 @@ public void serverError_UpdatePrimaryKeyWorkflow() throws Exception {
...
+		controllerTest.setEnablePrimaryKeyUpdate(false);
+		
+		try {
+			controllerTest.updatePrimaryKeyWorkflow("test", "test", "test", "test", "reason", response, mockAuth);
+			fail("Expected UnsupportedOperationException but got no exception.");
+		} catch(UnsupportedOperationException e) {
+			assertTrue(e.getMessage().contains("Feature not enabled"));
+		} catch(Exception e) {
+			fail("UnsupportedOperationException, but got " + e.getClass().getName());
+		}
```
cschroed-usgs, line 244: Not a blocker -- This test method carefully interleaves assertions at try-catch boundaries to test error behavior. JUnit 4 offers annotation-based exception assertions, and JUnit 5 offers a lambda-based inline assertion as alternatives. The JUnit 5 approach is pretty amenable to asserting things about the thrown exception's messages.
https://www.baeldung.com/junit-assert-exception

PR #139, opened 2020-05-27
Tweak email body based on feedback from LDMs.

zmoore-usgs:


src/main/java/gov/usgs/wma/mlrgateway/service/AdminService.java
```diff
@@ -66,13 +66,12 @@ public String generateSummaryHTML(String date, String bearerToken) throws IOExce
...
         if(summaryList != null && !summaryList.isEmpty()) {
-            builder.append("<p>This table shows the number of transactions that were executed which "); 
-            builder.append("<i>affected</i> the listed district code. This will not list transactions that ");
-            builder.append("resulted in no fields of a location being changed. This table does not infer ");
-            builder.append("that a given location still exists within a given district code. I.E: A ");
-            builder.append("modification of a location that moves it from one district code to ");
-            builder.append("another will be reflected in this table as a Location Modification for ");
-            builder.append("both the old and new district code.</p>");
+            builder.append("<p>The summary table below shows the daily number of executed transactions by "); 
+            builder.append("district code. The table does not list failed or duplication transactions. ");
```
kmschoep-usgs, line 70: should "duplication" be "duplicated"?  Otherwise, looks good.

zmoore-usgs: 🤷‍♂️ Joe and Kevin worked together to create the text and that's what they both agreed upon and came up with. I'm not sure if a "duplication transaction" is some sort of actual well-known LDM terminology for transactions that don't change anything?

kmschoep-usgs: Well then I guess it's all fine!

PR #143, opened 2020-06-12
Return parsable error message when no user email is available.

kmschoep-usgs:
use CC List in place of null recipient when no user email is available.

### DIFF
--- a/src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
+++ b/src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
@@ -25,9 +25,12 @@ import org.springframework.stereotype.Controller;
 public abstract class BaseController {
 	protected NotificationService notificationService;
 	protected UserAuthService userAuthService;
+	private String notificationEmailCCListString;
 	
 	@Value("${notification.email.cc-list:}")
-	private String notificationEmailCCListString;
+	protected void setNotificationEmailCCListString(String notificationEmailCCListString) {
+		this.notificationEmailCCListString = notificationEmailCCListString;
+	}
 
 	@Value("${uiHost:}")
 	protected String uiDomainName;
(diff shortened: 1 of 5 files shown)


src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -68,31 +71,35 @@ public static void remove() {
...
 				WorkflowController.addWorkflowStepReport(new StepReport(NotificationService.NOTIFICATION_STEP, status, false, ((FeignBadResponseWrapper) e).getBody()));
 			} else {
 				int status = HttpStatus.SC_INTERNAL_SERVER_ERROR;
-				WorkflowController.addWorkflowStepReport(new StepReport(NotificationService.NOTIFICATION_STEP, status, false, e.getMessage()));
+				WorkflowController.addWorkflowStepReport(new StepReport(NotificationService.NOTIFICATION_STEP, status, false, "{\"error_message\": \"" + e.getMessage() + "\"}"));
 			}
 		}
+		
+		userSummaryReport = userSummaryReportBuilder.buildUserSummaryReport(getReport());
+		notificationService.sendNotification(userEmail, ccList, notificationService.buildEmailSubject(subject), getReport().getUserName(), attachmentFileName, userSummaryReport);
```
zmoore-usgs, line 102: I think we'd still want these in the try/catch block, right? Otherwise the catch block is only catching errors from finding the user's email in the security context, and this being here means I think it will still run even if `userEmail` is empty (if it couldn't find the user email and the CCList was empty)

kmschoep-usgs: Right, and then the NotificationService will catch the error and throw a `Notification could not be sent.` error (which is not thrown when there is a CCList but no `userEmail`.  I found I needed to let the Notification service catch the error so that the lack of _any_ notification sent could be reported back to the user.

zmoore-usgs: Ahh gotcha, okay that makes sense.

PR #145, opened 2020-11-02
MLR-603: Bugfix; ddot submitted with lat/long to update but no datum

kmschoep-usgs:
code will be updated.

Also updated pom to push/pull to/from internal WMA Artifactory. URL.


src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -85,13 +88,25 @@ public void completeWorkflow(MultipartFile file) throws HystrixBadRequestExcepti
...
-					ml = legacyValidatorService.doValidation(ml, isAddTransaction, siteReport);
+					
+					//Fetch Existing Record
+					existingRecord = fetchExistingRecord(ml, isAddTransaction, siteReport);
+					
+					ml = legacyValidatorService.doValidation(ml, existingRecord, isAddTransaction, siteReport);
+					
+					// If this is an update, and latitude and longitude are being updated but the 
+					// datum is not included, pull that from the existing record
+					if (!isAddTransaction && ml.get(COORDINATE_DATUM_CODE) ==  null && ml.get(LATITUDE) != null && ml.get(LONGITUDE) != null ) {
```
zmoore-usgs, line 106: It feels a little out of place to me to have this field-specific logic here in the Gateway but I don't think there's a more logical place for this to happen. I was thinking perhaps as a transformation in the legacy transformer but then you'd have to pass the full existing record there anyways so it just ends up making the flow more complicated. So I think this works.

zmoore-usgs (approved):

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
kmschoep-usgs: Yeah, my original thought was going to be to pass in the existing record to the transformer, like the validator, but then would be modifying two services just to pass in one field.

SMH94 (approved):

pom.xml
```diff
@@ -26,17 +26,17 @@
...
             <id>wma-maven-centralized</id>
             <name>Water Mission Area Snapshots and Releases</name>
-            <url>https://artifactory.wma.usgs.gov/artifactory/wma-maven</url>
+            <url>https://artifactory.wma.chs.usgs.gov/artifactory/wma-maven</url>
         </repository>
         <repository>
             <id>eros-artifactory-wma-maven-releases</id>
             <name>EROS Artifactory</name>
-            <url>https://artifactory.wma.usgs.gov/artifactory/eros-artifactory-wma-maven-releases</url>
+            <url>https://artifactory.wma.chs.usgs.gov/artifactory/eros-artifactory-wma-maven-releases</url>
```
cschroed-usgs, line 39: Thank you!

src/main/java/gov/usgs/wma/mlrgateway/workflow/LegacyWorkflowService.java
```diff
@@ -240,6 +264,17 @@ public void updatePrimaryKeyWorkflow(String oldAgencyCode, String oldSiteNumber,
 		LOG.trace("End processing primary key update transaction [" + oldAgencyCode + "-" + oldSiteNumber + " to " + newAgencyCode + "-" + newSiteNumber + "] ");
 	}
 	
+	protected Map<String, Object> fetchExistingRecord(Map<String, Object> ml, Boolean isAddTransaction, SiteReport siteReport) {
+		Map<String, Object> existingRecord = new HashMap<>();
+		
+		//Fetch Existing Record
+		String siteNumber = ml.get(LegacyWorkflowService.SITE_NUMBER) != null ? ml.get(LegacyWorkflowService.SITE_NUMBER).toString() : null;
+		String agencyCode = ml.get(LegacyWorkflowService.AGENCY_CODE) != null ? ml.get(LegacyWorkflowService.AGENCY_CODE).toString() : null;
```
cschroed-usgs, line 272: It's not necessarily applicable here, but for learnsies, `Optional`s in Java 8+ are kinda cool and they can replace some null checks and/or ternary statements. https://www.baeldung.com/java-optional

cschroed-usgs (approved):
Solid work! Merge when ready.

PR #146, opened 2020-11-04
add new parameter in the config to store a list of URIs that the MLR

kmschoep-usgs:
auth is allowed to redirect to.

For AQS-PI and whatever other systems need to auth against MLR via a
browser.  Also upgraded some dependencies.


zmoore-usgs (approved):
Looks good to me!

SMH94 (approved):

SMH94:
Nice tests!

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
```diff
@@ -26,14 +26,17 @@
...
 	
 	@Value("${notification.email.cc-list:}")
 	protected void setNotificationEmailCCListString(String notificationEmailCCListString) {
 		this.notificationEmailCCListString = notificationEmailCCListString;
 	}
-
-	@Value("${uiHost:}")
-	protected String uiDomainName;
+	
+	@Value("${allowedRedirectUris:}")
```
cschroed-usgs, line 36: Untested:

It looks like the @Value annotation can convert CSV strings to Arrays of Strings.
https://www.baeldung.com/spring-inject-arrays-lists

Example...
```
@Value("${myListOfValues}")
private String[] valuesArray;
```


...or in this case something more like this?...

```
        protected String[] allowedRedirectUris;
        ...
	protected void setAllowedRedirectUriString(String[] allowedRedirectUris) {
		this.allowedRedirectUris = allowedRedirectUris;
	}
```

Would @Value's built-in CSV parsing work, or does parsing using StringUtils.split offer an advantage?

cschroed-usgs (approved):
Looks good to merge!

src/main/java/gov/usgs/wma/mlrgateway/controller/BaseController.java
cschroed-usgs: (this is not a blocker)

kmschoep-usgs: That's cool, and should work fine.

